PR #12, opened 2026-01-04
security: add CI workflow to build dist/ after merge

jarrodwatts:
## Summary

Phase 1 of removing dist/ from PRs to close a security attack vector.

**Problem**: PRs currently include compiled `dist/` files. A malicious contributor could inject code into `dist/` that doesn't match `src/`, and reviewers typically don't scrutinize compiled output.

**Solution**: CI builds and commits `dist/` to main after PR merges. This is Phase 1 - just adding the workflow.

## Changes

- Add `.github/workflows/build-dist.yml` - builds and commits dist/ on push to main
- Update `CONTRIBUTING.md` with build process documentation
- Reset `CHANGELOG.md` to 0.0.1 (initial release)

## What's Next (Phase 2)

After this PR is merged and we verify CI commits dist/ successfully:
1. Add `dist/` to `.gitignore`
2. Remove `dist/` from git tracking
3. Remove husky pre-commit hook

This two-phase approach ensures installation never breaks.

## Test plan

- [ ] Merge this PR
- [ ] Verify CI workflow runs and commits dist/ (or no-op if unchanged)
- [ ] Fresh plugin install still works

### DIFF
--- a/.github/workflows/build-dist.yml
+++ b/.github/workflows/build-dist.yml
@@ -27,6 +27,7 @@ jobs:
           cache: 'npm'
 
       - run: npm ci
+      - run: npm test
       - run: npm run build
 
       - name: Verify build output


jarrodwatts:
Thanks for the thorough review!

## Addressed

**Race Condition** - Fixed in 4b34cd9. Added `paths-ignore: ['dist/**']` to ci.yml so CI won't re-trigger when build-dist commits compiled output.

## Response to Other Points

**Branch Protection**: `GITHUB_TOKEN` has push access by default. If branch protection is later enabled with "require PR", we'd need to either:
- Add github-actions[bot] to bypass list, or  
- Use a PAT instead of GITHUB_TOKEN

For now, this works. Can adjust in Phase 2 if needed.

**Stale dist/**: The workflow will actually fail if `npm run build` fails (non-zero exit). The `git diff --staged --quiet ||` is just to avoid empty commits when dist/ is unchanged. Build failures will surface.

**CHANGELOG reset**: Confirmed - no existing git tags or releases. 0.0.1 is the true initial version.

Ready for merge when you are!

jarrodwatts:
Addressed in ef79fd6:

- **Concurrency**: Added `concurrency: group: build-dist` with `cancel-in-progress: false` to queue builds
- **Permissions**: Added explicit `permissions: contents: write`
- **Build verification**: Added `test -f dist/index.js || exit 1` step

Re: package.json version - it's already 0.0.1, aligned with the reset CHANGELOG.

PR #21, opened 2026-01-04
fix: correct context window size from 100k to 200k

yansircc:
## Summary
- Fix incorrect context percentage calculation when Claude Code reports 100k context window
- All modern Claude models (Opus 4.5, Sonnet, etc.) have 200k context windows
- When `context_window_size` is 100000, correct it to 200000

## Problem
Claude Code may report `context_window_size` as 100000 (100k), but all modern Claude models actually have 200k context windows. This causes the percentage to be doubled (e.g., 20% actual usage shows as 40%).

## Solution
Added a simple check: if `context_window_size === 100000`, use `200000` instead.

## Test plan
- [x] Verified the fix locally
- [x] 20% actual usage now correctly displays as 20%

### DIFF
--- a/src/stdin.ts
+++ b/src/stdin.ts
@@ -1,6 +1,9 @@
 import type { StdinData } from './types.js';
 import { AUTOCOMPACT_BUFFER } from './constants.js';
 
+// All modern Claude models have 200k context window
+const DEFAULT_CONTEXT_SIZE = 200000;
+
 export async function readStdin(): Promise<StdinData | null> {
   if (process.stdin.isTTY) {
     return null;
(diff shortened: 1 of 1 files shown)


jarrodwatts:
Thanks for the PR\! All current Claude models have 200k context windows, so I'm curious - in what scenario are you seeing Claude Code report 100k?

yansircc:
Hi @jarrodwatts, sorry for the late response!

I apologize for the confusion in my PR description. Looking back, I conflated two separate issues:

1. **The AUTOCOMPACT_BUFFER issue** (the real bug I encountered): The `getContextPercent` function adds `AUTOCOMPACT_BUFFER` (45k) to `totalTokens` before calculating the percentage, which inflates the displayed percentage. For example, if actual usage is 56k/200k = 28%, the HUD shows (56k+45k)/200k = 51%.

2. **The 100k window size issue** (what my PR tried to fix): This was a misdiagnosis.

Would you be open to a new PR that simply removes the `+ AUTOCOMPACT_BUFFER` from the calculation? The fix would be:
```diff
- return Math.min(100, Math.round(((totalTokens + AUTOCOMPACT_BUFFER) / size) * 100));
+ return Math.min(100, Math.round((totalTokens / size) * 100));
```

This would make the HUD percentage match the `/context` command output.

yansircc:
I've created a new PR with the correct fix: #43

This one simply removes the `+ AUTOCOMPACT_BUFFER` from the percentage calculation, which should make the HUD match the `/context` command output.

PR #22, opened 2026-01-04
feature: use bun + TypeScript when available for 4x faster startup

jmherbst:
## Summary

- Auto-detects bun and uses it with TypeScript source directly
- Falls back to node + compiled JS when bun not available
- Bun's native TS transpiler is faster than loading pre-compiled JS

## Benchmarks

| Runtime | Average | Speedup |
|---------|---------|---------|
| bash -c + node + JS (before) | 89ms | baseline |
| bash -c + bun + TS (after) | 22ms | **4x faster** |

(10 runs each, macOS)

## Test plan

- [x] Verified command works with bun available (uses bun + src/index.ts)
- [x] Verified fallback to node + dist/index.js when bun not in PATH

### DIFF
--- a/commands/setup.md
+++ b/commands/setup.md
@@ -9,12 +9,14 @@ Add this statusLine configuration to `~/.claude/settings.json`:
 {
   "statusLine": {
     "type": "command",
-    "command": "bash -c 'node \"$(ls -td ~/.claude/plugins/cache/claude-hud/claude-hud/*/ 2>/dev/null | head -1)dist/index.js\"'"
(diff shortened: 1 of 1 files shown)


jmherbst:
Looks like this is being included in #24 🍻 cheers

jarrodwatts:
Thanks for the contribution @jmherbst , yes trying to support any platform more broadly by giving some better setup instructions - should be merged soon!

jarrodwatts:
Works with Bun + TS now @jmherbst !
image

Issue #29, opened 2026-01-05
feat: Support weekly usage limit display on status bar


jellydn:
## What
Display the weekly API usage limit on the Claude HUD status bar, showing remaining tokens/requests for the current billing period.

## Why
**User visibility**: Users currently can't see their weekly usage consumption in real-time. This leads to:
- Unexpected limit errors mid-work
- Difficulty budgeting usage across the week
- No indication when approaching limits
- Uncertainty about when limits reset

**Better UX**: Showing usage proactively:
- Helps users plan their work
- Provides warning before hitting limits
- Reduces frustration from abrupt cutoffs
- Aligns with existing context monitoring (we already show context window usage)

### Data Source
The weekly limit data needs to come from Claude's API configuration. Likely available via:
- API endpoint: Check if Claude Code exposes usage stats in settings or API
- Local cache: May be stored in `~/.claude/` configuration
- stdin JSON: Add to existing `StdinData` interface (ideal if available)

Suggested fields:
```typescript
interface StdinData {
  // ... existing fields
  usage_limits?: {
    weekly?: {
      limit: number;           // Total weekly allowance
      used: number;            // Tokens/requests used this week
      reset_at: string;        // ISO timestamp of reset
    }
  }
}
```

### Implementation Plan

1. **Update type definitions** (`src/types.ts`)
   - Add `UsageLimits` interface
   - Extend `StdinData` to include optional `usage_limits` field

2. **Parse usage data** (`src/stdin.ts`)
   - Extract usage limits from stdin JSON
   - Calculate percentage remaining
   - Format reset time (e.g., "resets in 2 days")

3. **Render on status line** (`src/render/session-line.ts`)
   - Add usage indicator to session line: `📊 78% week used`
   - Color coding: 
     - 🟢 <70%: green
     - 🟡 70-85%: yellow  
     - 🔴 >85%: red (show exact counts)
   - Conditionally show (only if limits data available)

4. **Handle edge cases**
   - Missing data: Don't display (graceful degradation)
   - Free tier vs paid: Different limits
   - Multiple models: May need separate tracking per model

### Example Output
```
[Opus] ████████░░ 45% | 📊 78% week used (resets in 2d) | 📋 3 rules | ⏱️ 12m
```

When approaching limit:
```
[Opus] ████████░░ 45% | 🔴 92% week used (1.2M/1.3M remaining) | 📋 3 rules | ⏱️ 12m
```

### Open Questions
[...]

PR #32, opened 2026-01-05
feat: configurable path levels (1-3) and git status toggle

Tsopic:
## Summary

Full HUD customization with configurable options:

1. **Layout styles** — Choose between `default`, `condensed`, or `separators` layouts
2. **Configurable path levels** — Show 1, 2, or 3 directory segments (default: 1)
3. **Git status toggle** — Enable/disable git branch display
4. **Dirty indicator** — Show `*` for uncommitted changes (default: on)
5. **Ahead/behind counts** — Show `↑N ↓N` for remote tracking (default: off)
6. **Display options** — Toggle visibility of all HUD elements
7. **Interactive CLI** — Configure via `npx claude-hud-configure`
8. **Menu-based navigation** — Jump to any section, go back anytime
9. **Live preview** — See HUD changes in real-time as you configure
10. **JSON config** — Or edit `~/.claude/plugins/claude-hud/config.json` directly

## Layout Options

**Default layout** — Everything on first line:
```
📁 my-project git:(main) | [Opus] ████░░░░░░ 42% | 2 rules | ⏱️ 5m
✓ Read ×3 | ✓ Edit ×1
```

**Condensed layout** — Model/context on top, project on bottom:
```
[Opus] ████░░░░░░ 42% | ⏱️ 5m
✓ Read ×3 | ✓ Edit ×1
📁 my-project git:(main) | 2 rules
```

**Separators layout** — Condensed with visual separators:
```
[Opus] ████░░░░░░ 42% | ⏱️ 5m
──────────────────────────────────────────────────
✓ Read ×3 | ✓ Edit ×1
──────────────────────────────────────────────────
📁 my-project git:(main) | 2 rules
```

## Interactive CLI with Menu Navigation

The configure tool now has a **main menu** that lets you jump to any section:

```
=== Claude HUD Configuration ===

✓ Existing configuration found

── HUD Preview (condensed) ──
  ███████░░░ 72% | ⏱️  5m
  ◐ Edit: .../auth.ts | ✓ Read ×3
  ✓ explore [haiku]: Finding auth code (3s)
  ▸ Write unit tests (1/3)
  📁 projects/apps/my-project git:(main) | 1 CLAUDE.md | 2 rules | 5 MCPs | 3 hooks

─── Main Menu ───
? What would you like to configure?
❯ Layout          (Condensed)
  Path Display    (3 levels)
  Git Status      (on, dirty)
  Session Line    (4/5 options)
  Additional Lines (3/3 lines)
  ✓ Save & Exit
[...]

PR #34, opened 2026-01-05
feat: add usage limits display for Pro/Max/Team users

melon-hub:
## Summary

  Adds opt-in display of 5-hour and 7-day rate limit consumption in the statusline for Pro/Max/Team subscribers.

  - Shows `5h: X% | 7d: Y%` when enabled
  - Shows limit reached warning with reset countdown
  - Uses existing OAuth credentials (zero token cost)
  - Graceful degradation when API unavailable

  **Normal usage:**
  📁 my-project git:(main) | [Opus 4.5 | Max] ████░░ 45% | 5h: 23% | 7d: 45%

  **When limit reached:**
  📁 my-project git:(main) | [Opus 4.5 | Max] ████░░ 45% | ⚠ Limit reached (resets 2h 15m)

  ## Background & Research

  This addresses the long-standing request in #29. During investigation, I discovered:
 1. **Credentials location**: Claude Code stores OAuth tokens in `~/.claude/.credentials.json` with structure including `claudeAiOauth.accessToken` and `subscriptionType`

  2. **Undocumented API**: Found `api.anthropic.com/api/oauth/usage` endpoint that returns utilization percentages (0-100) for both 5-hour and 7-day windows, plus reset timestamps

  3. **Zero cost**: The endpoint uses the existing OAuth token - no additional API calls or token consumption

  ## Design Decisions

  | Decision | Rationale |
  |----------|-----------|
  | **Opt-in only** | Undocumented API may change; users should consciously enable |
  | **60-second cache** | Balance freshness vs API load |
  | **`usage: ⚠` indicator** | Users know when API fails (can check DEBUG logs) |
  | **Graceful fallback** | API users see no usage display (they don't have rate limits) |

  ## Security Considerations

  - Token only sent to `api.anthropic.com` (hardcoded)
  - Token never logged in debug output
  - Read-only access to credentials file
  - Same token Claude Code already uses

  ## API Users

  This feature is for **Pro/Max/Team subscribers only**. API users are automatically excluded because:

  - API users pay per token, not via rate-limited subscriptions
  - They don't have 5-hour/7-day usage windows to display
  - The `subscriptionType` field in credentials identifies them

  When an API user has `CLAUDE_HUD_SHOW_USAGE=1` set, the feature silently does nothing (no error, no display). 

  ## Test plan

  - [x] 81 tests pass (15 new for usage API, 6 for rendering)
  - [x] Works with Pro/Max/Team subscriptions
  - [x] Graceful fallback for API users (no display)
  - [x] Handles API failures with warning indicator
  - [x] Edge cases: NaN, Infinity, negative values, expired tokens

  Supports #29


melon-hub:
Thought this was a neat opt-in idea which seems to work nicely locally, let me know what you think! @jarrodwatts 

image

jarrodwatts:
Oh wow - I wanted to add this but wasn't sure if it was doable... undocumented api is very interesting

Only feedback is I feel like we are showing a lot of info on the first line now (since we also merged the git branch + project name in there today)

Maybe we should consider breaking some parts into a new line  + also want to make it configurable to what shows and what doesn't in a config file somewhere ideally 

I'll ask claude to review

jarrodwatts:
@claude review PR ultrathink

jarrodwatts:
breh

melon-hub:
You make a really valid point, I'm not knowledgeable in this area but the research I did seem to just say people prefer not to but no real security risk (I'll defer to experts), so I thought opt-in perhaps if someone really wants it. 

I agree with you re number of lines and customization, it becomes a lot I really like the picture I just shared above simple but effective but I was thinking you could try and design like 1-3 primary options, and then allow some configuration if people want their very specific view? I saw in this #https://github.com/jarrodwatts/claude-hud/pull/32 that 1-3 configurable path levels which is could be neat. 

**I found a way as well to de-clutter it with that timed information, i.e only show 7d consumption when >80%**
image

**Moved some items to bottom row which change irregularly**
image

**Might be a bit much but I made this version to** - Line length dynamic to length of top line + 5 char
image

Reading a bit more it seems possible to build some of those tweaks as slash commands to the status like via the settings.json

I didn't implement those other styles but feel free to play around with what I started if you find this useful - I'm keen to try help with some settings options,

melon-hub:
Ok last contribution for the night - I just added env variables for layout customization and updated the PR. Will leave it with you!

**Layout Options (ENV Variable)**
```bash
export CLAUDE_HUD_LAYOUT=default     # Primary Layout (all on line 1)
export CLAUDE_HUD_LAYOUT=condensed   # Split layout, no separators
export CLAUDE_HUD_LAYOUT=separators  # Split layout with lines
```

jarrodwatts:
#35

Issue #44, opened 2026-01-07
resetAt.getTime is not a function


martinambrus:
## Summary
With newest version of both, Claude Code CLI and claude-hud, I started to get this error on WSL: [claude-hud] Error: resetAt.getTime is not a function

## Steps to Reproduce
Install latest versions, use Claude Code.

## Expected Behavior
Status line should display HUD

## Actual Behavior
Error appears

## Environment

- OS: Windows 11, WSL2
- Node/Bun version: Node v25.2.0
- Claude Code version: 2.0.76

## Logs or Screenshots
-

quocdatPTIT:
I have the same issue on Fedora

Goldfive:
@martinambrus @quocdatPTIT This PR fixes it: https://github.com/jarrodwatts/claude-hud/pull/45

jarrodwatts:
Should be fixed thanks to @Goldfive

quocdatPTIT:
Hi @jarrodwatts, @Goldfive. When I start a Claude Code session, the Claude Hub appears correctly, but the bug still occurs as soon as I send a message.

<img width="1396" height="582" alt="Image" src="https://github.com/user-attachments/assets/da90204a-b3cb-45e0-85c1-2f974002eb87" />

<img width="1396" height="582" alt="Image" src="https://github.com/user-attachments/assets/71e732d4-fab5-4e3a-bdef-8461a1c1af94" />

martinambrus:
I'm afraid this is still happening for me with the 0.0.4 release on Claude Code 2.1.1 as well. I also vote for reopening this.

PR #45, opened 2026-01-07
fix: reconvert Date objects when reading from cache

Goldfive:
## Summary

Fixes the `resetAt.getTime is not a function` error reported in #44.

## Root Cause

`JSON.stringify()` converts `Date` objects to ISO strings (e.g., `"2024-01-15T12:00:00.000Z"`), but `JSON.parse()` doesn't automatically convert them back to `Date` objects. 

When `readCache()` returns cached data, the `fiveHourResetAt` and `sevenDayResetAt` fields are strings instead of `Date` objects, causing `formatResetTime()` to fail when calling `.getTime()` on them.

## Fix

Reconvert the date strings back to `Date` objects after parsing the cache file. The `new Date()` constructor safely handles both Date objects and ISO strings, so this works regardless of whether the data came from cache (string) or fresh API response (Date).

## Test plan

- [x] All existing tests pass (110/110)
- [x] Manually verified HUD works after clearing cache

### DIFF
--- a/src/usage-api.ts
+++ b/src/usage-api.ts
@@ -61,7 +61,17 @@ function readCache(homeDir: string, now: number): UsageData | null {
     const ttl = cache.data.apiUnavailable ? CACHE_FAILURE_TTL_MS : CACHE_TTL_MS;
     if (now - cache.timestamp >= ttl) return null;
 
-    return cache.data;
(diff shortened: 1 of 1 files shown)


jarrodwatts:
Thanks a lot!

JensMyParcel:
> Thanks a lot!

Thank Claude :P

PR #49, opened 2026-01-07
Fix HUD context percentage to match /context across window sizes

### LINKED ISSUE #48: AUTOCOMPACT_BUFFER causes percentage mismatch with /context command
## Summary

The HUD displays ~38-40% context usage when `/context` shows only 16%. The discrepancy is caused by the 45k `AUTOCOMPACT_BUFFER` added in `src/stdin.ts`.

## Details

**constants.ts:5**
```typescript
export const AUTOCOMPACT_BUFFER = 45000;
```

**stdin.ts:35-40**
```typescript
const totalTokens =
  (usage.input_tokens ?? 0) +
  (usage.cache_creation_input_tokens ?? 0) +
  (usage.cache_read_input_tokens ?? 0);

return Math.round(((totalTokens + AUTOCOMPACT_BUFFER) / size) * 100);
```

## Example

- `/context` reports: 31k/200k = **16%**
- HUD calculates: (31k + 45k) / 200k = 76k/200k = **38%**

## Comment contradiction

The comment in `constants.ts` says:
> "must be added to match /context output"

But it actually causes the opposite - it makes the HUD NOT match `/context`.

## Questions

1. Is this buffer still needed?
2. Has the `/context` calculation changed since this was implemented?
3. Should the buffer be removed, or should `/context` be the one that's "wrong"?

## Environment

- Claude Code version: latest
- Model: claude-opus-4-5-20251101
- OS: macOS

StephenJoshii:
## Summary
Fixes #48

Fix context percentage mismatch between HUD and /context command for context windows larger than 200k tokens.

The autocompact buffer was hardcoded as 45000 tokens, which only worked correctly for 200k windows (45k = 22.5% of 200k). For larger windows (500k, 1M), the HUD showed significantly lower percentages than /context.

Changed the buffer from a fixed value to a dynamic 22.5% of the context window size:

// Before
export const AUTOCOMPACT_BUFFER = 45000;

// After  
export const AUTOCOMPACT_BUFFER_PERCENT = 0.225;
const autocompactBuffer = size * AUTOCOMPACT_BUFFER_PERCENT;

Window | Before | After | /context
200k | 45% | 45% | 45% 
500k | 23% | 36% | 36% 
1M | 9% | 27% | 27% 

## Testing

- [ ] `npm test`
- [ ] `npm run test:coverage`

## Checklist


StephenJoshii:
simply changed the hardcoded 45000 buffer to a dynamic: windowSize * 0.225.

jarrodwatts:
Thanks @StephenJoshii - but what claude models have context windows that are not 200k?

saw this issue raised once before but wasn't sure how people run into it

StephenJoshii:
Larger enterprise/ preview models, or any setup where the context size is set differently. In those cases the hardcoded value stops matching the intended behavior.

jarrodwatts:
Thank you for the percentage-based approach! This was incorporated into #55 - the buffer is now 22.5% of context window size instead of hardcoded 45k, so it scales correctly for enterprise windows.

Credit: Your idea of using a percentage instead of fixed tokens was key to the solution. 🙏

Issue #70, opened 2026-01-13
Can't install


rzgrozt:
## Summary
╭─    ~ ··········································· ✔  took 32s   at 12:44:52 ÖS  
╰─ claude plugin install claude-hud
Installing plugin "claude-hud"...
✘ Failed to install plugin "claude-hud": EXDEV: cross-device link not permitted, rename '/home/ruzgar/.claude/plugins/cache/claude-hud' -> '/tmp/claude-plugin-temp-1768297496499'
## Environment

- OS: CachyOS
- Node/Bun version: bun 1.3.5
- Claude Code version: 2.1.6

jarrodwatts:
Same root cause as #52 - `/tmp` is on a different filesystem (tmpfs) than `~/.claude`.

**Workaround**:
```bash
mkdir -p ~/.cache/tmp && TMPDIR=~/.cache/tmp claude /plugin install claude-hud
```

Then run `/claude-hud:setup` as normal.

#74 adds a visible warning in the README and improves the setup command to detect this issue proactively.

jarrodwatts:
Fixed in #75 - README now has Linux warning, setup detects EXDEV issue and provides workaround.

PR #74, opened 2026-01-14
fix: move EXDEV detection to run first on Linux setup

### LINKED ISSUE #52: Failed to install: EXDEV: cross-device link not permitted, rename '/home/mi/.claude/plugins/cache/claude-hud' -> '/tmp/claude-plugin-temp-1767864145652'
## Summary
Error: Failed to install: EXDEV: cross-device link not permitted, rename '/home/mi/.claude/plugins/cache/claude-hud' -> '/tmp/claude-plugin-temp-1767864145652'

## Environment

- OS:
- Node/Bun version:
- Claude Code version:

jarrodwatts:
## Summary

- Fixes the EXDEV detection that was added in #53 but placed in an unreachable location
- Adds visible Linux warning in README before users attempt installation
- Restructures setup.md to check for cross-device filesystem issue FIRST on Linux

## Problem

PR #53 added EXDEV detection inside the "plugin path empty" check in setup.md. This was unreachable because:

1. User runs `/plugin install claude-hud`
2. Installation fails with EXDEV error
3. User can't run `/claude-hud:setup` because plugin isn't installed
4. The detection code never executes

Users in #52 and #70 confirmed the fix didn't help.

## Solution

**README.md**: Collapsible warning before install step catches users BEFORE they hit the error.

**setup.md**: New "Step 0" runs EXDEV check unconditionally on Linux:
- Runs FIRST, before checking plugin path
- Uses `~` instead of `~/.claude` (works even if .claude doesn't exist yet)
- If CROSS_DEVICE detected AND plugin not installed → shows workaround
- If plugin already installed → continues normally

## Test plan

- [ ] Linux user with tmpfs `/tmp` sees README warning before installing
- [ ] `/claude-hud:setup` on Linux runs Step 0 check first
- [ ] CROSS_DEVICE detection works with fresh install (no ~/.claude yet)
- [ ] macOS/Windows users unaffected (Step 0 skipped)

Fixes #52, #70

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,24 @@ Inside a Claude Code instance, run the following commands:
 ```
 
 **Step 2: Install the plugin**
+
+<details>
+<summary><strong>⚠️ Linux users: Click here first</strong></summary>
+
+On Linux, `/tmp` is often a separate filesystem (tmpfs), which causes plugin installation to fail with:
+```
+EXDEV: cross-device link not permitted
+```
+
+**Fix**: Set TMPDIR before installing:
+```bash
+mkdir -p ~/.cache/tmp && TMPDIR=~/.cache/tmp claude
+```
+
+Then run the install command below in that session. This is a [Claude Code platform limitation](https://github.com/anthropics/claude-code/issues/14799).
+
+</details>
+
 ```
 /plugin install claude-hud
 ```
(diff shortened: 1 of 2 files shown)


jarrodwatts:
Closing in favor of #75 which includes this EXDEV fix plus ghost installation detection. No need to merge both.

PR #93, opened 2026-01-19
fix(setup): ask user for shell type on Windows instead of assuming Git Bash

### LINKED ISSUE #90: Bug: Incorrectly Detects Git Bash on PowerShell Systems
# Setup Detection Bug: Incorrectly Detects Git Bash on PowerShell Systems

  ## Problem

  The `/claude-hud:setup` skill incorrectly detects Git Bash/MSYS2 environment when users are running PowerShell (including PowerShell 7/pwsh), causing setup to fail.

  ## Root Cause

  The detection logic in **Step 1** uses:
  ```bash
  uname -s
  ```

  This command detects the Bash tool's underlying environment, not the user's actual shell. On Windows, Claude Code's Bash tool often uses Git Bash/MSYS2 as its interpreter, so uname -s returns MINGW64_NT-... even when the user is running PowerShell 7.

  This causes the setup to:
  1. Detect MINGW64_NT-...
  2. Follow the "macOS/Linux/Git Bash" instruction path
  3. Generate bash-style commands like:
  ```bash
  bash -c '"node" "$(ls -td ~/.claude/plugins/cache/claude-hud/claude-hud/*/ | head -1)dist/index.js"'
  ```
  4. These commands fail in PowerShell environments

  Steps to Reproduce

  1. Be on Windows with PowerShell 7 installed
  2. Run /claude-hud:setup
  3. The skill detects MINGW64_NT-... from uname -s
  4. Setup attempts to use bash-style commands
  5. Commands fail in PowerShell environment

  Expected Behavior

  The skill should:
  - Detect that the user is on Windows PowerShell
  - Use PowerShell commands (pwsh -Command ...)
  - Successfully complete setup

  Actual Behavior

  The skill:
  - Detects Git Bash environment (false positive)
  - Uses bash-style commands
  - Fails to complete setup
  - User must manually correct and specify "I'm on PowerShell"

  Impact

  PowerShell users (likely the majority of Windows users) experience failed setups and must manually intervene.

  Suggested Fix

  Update the detection logic in Step 1 to check multiple signals:
```bash
  # Better detection approach
  # 1. Check if we're on Windows first (working directory format)
  if [[ "$PWD" =~ ^/[a-zA-Z]/ ]] || [[ "$PWD" =~ ^[A-Z]:\\ ]]; then
    # We're on Windows - default to PowerShell unless user is in actual Git Bash session
    # Check if SHELL is bash and user explicitly wants bash-style commands
    if [[ "$SHELL" =~ bash ]] && [[ ! "$TERM_PROGRAM" =~ "vscode" ]]; then
      # User is in native Git Bash/WSL - use bash commands
      USE_POWERSHELL=false
    else
      # Use PowerShell (covers PowerShell 7, Windows Terminal, VSCode, etc.)
      USE_POWERSHELL=true
    fi
  else
    # macOS/Linux
    USE_POWERSHELL=false
  fi
```

  Or simpler: **Ask the user which shell they're using if on Windows**:
```bash
  - "Detected Windows. Are you using: PowerShell / Git Bash / WSL?"
```

  Environment

  - OS: Windows 11 23H2
  - Shell: PowerShell 7.5.4
  - Claude Code 2.1.12
  - claude-hud: 0.0.6

  Workaround

  Users must manually tell Claude "I'm on PowerShell" to force correct detection.

  Related

  This affects the setup skill instructions, specifically the platform detection logic that determines which command format to generate.

jarrodwatts:
## Summary

- When `uname -s` returns MINGW/MSYS/CYGWIN, the setup now asks the user which shell they're actually using via `AskUserQuestion`
- Users who select PowerShell are directed to the Windows (PowerShell) instructions
- Users who select Git Bash/MSYS2/Cygwin continue with the bash-style commands as before
- Added a detection logic table and explicit guidance for the agent to follow

## Root Cause

The previous detection logic assumed that if `uname -s` returned a MINGW/MSYS/CYGWIN variant, the user was using Git Bash. However, on Windows, Claude Code's Bash tool always uses Git Bash/MSYS2 as its interpreter regardless of the user's actual shell environment. This caused PowerShell users (the majority of Windows users) to receive bash-style commands that failed.

## Test plan

- [ ] Run `/claude-hud:setup` on Windows with PowerShell - should ask which shell you're using
- [ ] Select "PowerShell" - should follow PowerShell instructions and generate PowerShell commands
- [ ] Run `/claude-hud:setup` on Windows with actual Git Bash - should ask which shell and selecting Git Bash should use bash commands
- [ ] Run `/claude-hud:setup` on macOS/Linux - should continue working as before without asking

Fixes #90

### DIFF
--- a/commands/setup.md
+++ b/commands/setup.md
@@ -91,9 +91,34 @@ This is a [Claude Code platform limitation](https://github.com/anthropics/claude
 
 ## Step 1: Detect Platform & Runtime
 
-**macOS/Linux** (if `uname -s` returns "Darwin", "Linux", or a MINGW*/MSYS*/CYGWIN* variant):
+First, detect the platform:
+```bash
+uname -s
+```
 
-> **Git Bash/MSYS2/Cygwin users on Windows**: Follow these macOS/Linux instructions, not the Windows section below. Your environment provides bash and Unix-like tools.
+### Platform Detection Logic
+
+| `uname -s` result | Platform | Next Step |
+|-------------------|----------|-----------|
+| Darwin | macOS | Follow **macOS/Linux** instructions |
+| Linux | Linux (native or WSL) | Follow **macOS/Linux** instructions |
+| MINGW* / MSYS* / CYGWIN* | Windows (Git Bash detected) | **IMPORTANT**: Ask user which shell they use |
+| Command fails / not found | Windows (PowerShell) | Follow **Windows (PowerShell)** instructions |
+
+**CRITICAL for Windows detection (MINGW/MSYS/CYGWIN)**: When `uname -s` returns a MINGW/MSYS/CYGWIN variant, this means Claude Code's Bash tool is using Git Bash - but **this does NOT mean the user is actually using Git Bash as their shell**. Most Windows users use PowerShell. You **MUST** ask the user:
+
+Use AskUserQuestion:
+- Question: "You're on Windows. Which shell are you using?"
+- Options: "PowerShell (or Windows Terminal)" / "Git Bash, MSYS2, or Cygwin"
+
+**If PowerShell**: Follow the **Windows (PowerShell)** instructions below.
+**If Git Bash/MSYS2/Cygwin**: Follow the **macOS/Linux** instructions (these bash-based environments work like Unix).
+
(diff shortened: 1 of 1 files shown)


jarrodwatts:
Closing in favor of a better approach - using Claude's Platform context instead of asking users.

Issue #96, opened 2026-01-19
Feature Request: Add sonnet limit [weekly?] claude code has


brettstark73:
## Summary
Feature Request: Add sonnet limit [weekly?] claude code has

## Problem to Solve
see where you are with respect to the weekly sonnet only limit.

## Proposed Solution
i could probably do it with Claude Code, but hit the limit :(

jarrodwatts:
Addressed via the new  config (merged in #107). Set it to  to always show the 7‑day usage, which covers the weekly Sonnet limit visibility. Closing as resolved.

jarrodwatts:
Added a weekly (7d) usage threshold so you can surface the Sonnet weekly limit. Set `display.sevenDayThreshold` to `0` to always show the 7-day usage (default is 80%). This is merged (PR #107). Closing out as resolved.

PR #107, opened 2026-02-03
feat(config): add seven-day usage threshold

### LINKED ISSUE #98: Feature Request: Add config option for 7-day usage display threshold
## Summary

Allow users to configure the threshold at which 7-day usage is displayed, instead of the hardcoded 80%.

## Problem to Solve

Currently, 7-day usage is only shown when `sevenDay >= 80`:

```typescript
// src/render/lines/usage.ts:42
if (sevenDay !== null && sevenDay >= 80) {
```

This means users cannot see their weekly usage until they've consumed 80% of their 5-hour limit. Many users would prefer to always see both metrics to better track their usage patterns.

## Proposed Solution

Add a new config option `sevenDayThreshold` to control when 7-day usage appears:

```json
{
  "display": {
    "showUsage": true,
    "sevenDayThreshold": 0
  }
}
```

- `0` = always show 7-day usage
- `80` = current behavior (default)
- Any value 0-100 = show when 7-day usage >= threshold

## Alternatives Considered

- Boolean flag `alwaysShowSevenDay: true/false` - simpler but less flexible
- Remove the threshold entirely - but some users may prefer the current behavior

## Additional Context

Related to #96 (sonnet weekly limit visibility).

Currently requires manually editing the compiled JS file as a workaround:
```bash
sed -i '' 's/sevenDay >= 80/sevenDay >= 0/g' ~/.claude/plugins/cache/claude-hud/claude-hud/0.0.6/dist/render/lines/usage.js
```

jarrodwatts:
## Summary
- add display.sevenDayThreshold config to control when 7-day usage appears
- update usage rendering and docs/tests to respect the threshold

## Testing
- npm test

Closes #98

Related: #96


devseunggwan:
@jarrodwatts thanks for this PR🙌

PR #128, opened 2026-02-19
fix: support win32 + bash shell in setup command

### LINKED ISSUE #121: /claude-hud:setup generates broken command on Windows (win32)
# Issue: `/claude-hud:setup` generates broken command on Windows (win32)

## Problem

On Windows (`Platform: win32`), the setup command generates a statusLine command that has two issues:

1. **PowerShell window flashing**: The generated command wraps execution in `powershell -Command "..."`, which spawns a visible console window every ~300ms when Claude Code refreshes the statusLine. This causes constant black window flickering on screen.

2. **`bun` resolves to a shell wrapper, not an executable**: On Windows (e.g. via nvm4w), `bun` at `D:\nvm4w\nodejs\bun` is actually a bash shell script, not a `.exe`. PowerShell cannot execute it. The real binary is at `D:\nvm4w\nodejs\node_modules\bun\bin\bun.exe`.

3. **Marketplace update overwrites manual fix**: After manually fixing the statusLine command, the plugin system periodically re-scans marketplaces and regenerates the default (broken) command, overwriting the user's fix. This typically happens ~24 hours later, so the fix appears to work same-day but breaks the next day.

## Expected Behavior

The setup command should:
- Detect `win32` platform and generate a direct executable command (no PowerShell wrapper) to avoid window flashing
- Resolve `bun`/`node` to the actual `.exe` binary path, not a shell wrapper
- Not overwrite a user-customized statusLine command on marketplace refresh

## Current Workaround

Manually set statusLine in `~/.claude/settings.json` with a direct bun.exe path:

```json
{
  "statusLine": {
    "type": "command",
    "command": "D:\\nvm4w\\nodejs\\node_modules\\bun\\bin\\bun.exe C:\\Users\\<user>\\.claude\\plugins\\cache\\claude-hud\\claude-hud\\0.0.7\\src\\index.ts"
  }
}
```

And use a self-repair script triggered via CLAUDE.md to auto-fix after overwrites.

## Environment

- OS: Windows 10+ (MSYS_NT / Git Bash)
- Claude Code: 2.1.39
- claude-hud: 0.0.7
- Runtime: Bun (installed via nvm4w)
- Claude Code shell: Git Bash (but statusLine executes via Windows native process)

sanghyun-io:
## Summary

Fixes #121

- Add shell detection in `commands/setup.md` Step 1 for `win32` platform
- When Shell is `bash` (Git Bash/MSYS2), redirect to macOS/Linux bash instructions instead of generating PowerShell commands
- Existing PowerShell/cmd behavior is unchanged

## Root Cause

Claude Code executes the `statusLine` command using the user's **detected shell** (shown as `Shell:` in the environment context). On Git Bash environments, `Platform: win32` but `Shell: bash`. The previous setup logic only checked `Platform` and always generated a PowerShell command for `win32`, creating a 3-layer process chain:

```
bash -c 'powershell -Command "& { ... bun ... }"'
```

This chain fails because:
1. `$` variables in the PowerShell command are interpreted by bash first
2. PowerShell startup adds ~2s latency, risking the statusLine timeout
3. stdin pipe (JSON data) breaks across the bash→PowerShell boundary

The fix detects the shell and uses the native bash command format when the shell is bash, avoiding the cross-shell chain entirely.

## Changes

**`commands/setup.md`**:
- Updated platform table to include Shell column with `win32 + bash` row
- Added shell detection guidance at top of Windows section
- Existing PowerShell instructions wrapped under conditional heading

## Test Plan

- [x] Verified bash command works on MINGW64 (Git Bash): `echo '{"model":{"id":"test"}}' | bash -c '"bun" "$(ls -td ~/.claude/plugins/cache/claude-hud/claude-hud/*/ | head -1)src/index.ts"'`
- [ ] Verify PowerShell path is unaffected on native Windows PowerShell
- [ ] Verify macOS/Linux paths are unaffected

### DIFF
--- a/commands/setup.md
+++ b/commands/setup.md
@@ -93,11 +93,12 @@ This is a [Claude Code platform limitation](https://github.com/anthropics/claude
 
 **IMPORTANT**: Determine the platform from your environment context (`Platform:` value), NOT from `uname -s`. The Bash tool may report a different environment than the user's actual platform (e.g., Git Bash on Windows reports MINGW even when the user launched Claude Code from PowerShell).
 
-| Platform | Command Format |
-|----------|---------------|
-| `darwin` | bash (macOS) |
-| `linux` | bash (all Linux distros including NixOS, Ubuntu, Arch, etc.) |
-| `win32` | PowerShell (works universally on Windows 10+) |
+| Platform | Shell | Command Format |
+|----------|-------|---------------|
+| `darwin` | any | bash |
+| `linux` | any | bash |
+| `win32` | `bash` (Git Bash, MSYS2) | bash (same as macOS/Linux) |
+| `win32` | PowerShell / cmd | PowerShell |
 
 ---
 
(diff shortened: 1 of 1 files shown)


jarrodwatts:
Thanks for the contribution. I merged this fix path in #164 with additional validation and related regressions bundled in a single maintainer pass, so I’m closing this as superseded.

PR #130, opened 2026-02-20
feat: add AI-powered session summary to statusline

sangmandu:
## Summary

When running multiple Claude Code terminals simultaneously, it's difficult to tell which session is working on what. This PR adds an **opt-in session summary** feature that uses Haiku (via the Claude Agent SDK) to generate a 2-line summary displayed in the statusline.

**Example output:**
```
─── session summary ───
🏷️ claude-hud plugin — adding Haiku session summary
📍 Background worker implemented, testing cache
```

## Architecture

The key challenge is that statusline rendering must be fast (<300ms), but an AI API call takes seconds. This is solved with a **non-blocking background worker** pattern:

1. **Main render loop** (`summary.ts`) — only reads a cached JSON file. If the cache is stale (based on `summaryInterval` turns), it spawns a background worker and returns the old cached summary.
2. **Background worker** (`summary-worker.ts`) — a detached Node.js process that calls Haiku via the Agent SDK, then writes the result to a per-session cache file.
3. **Per-session isolation** — cache and lock files are keyed by MD5 hash of the transcript path, so multiple terminals don't interfere with each other.

### Why Agent SDK instead of direct API?

The Agent SDK handles OAuth authentication automatically (no API key management needed), making it work seamlessly for Claude Code users who are already authenticated.

### Why Node.js worker instead of in-process?

The Agent SDK spawns a Claude Code subprocess internally, which is incompatible with running inside bun (which is how the plugin's `src/` files are executed). The worker runs as a compiled `dist/summary-worker.js` via Node.js to avoid this.

## Configuration

```json
{
  "display": {
    "showSummary": true,
    "summaryInterval": 5
  }
}
```

| Option | Type | Default | Description |
|--------|------|---------|-------------|
| `showSummary` | boolean | `false` | Enable session summary |
| `summaryInterval` | number | `5` | Refresh every N user turns (1–20) |

## Files changed

| File | Change |
|------|--------|
| `src/summary.ts` | **New** — Orchestration: cache read, worker spawn, lock management |
| `src/summary-worker.ts` | **New** — Background Haiku call via Agent SDK |
| `src/render/summary-line.ts` | **New** — Render summary lines |
| `src/types.ts` | Added `SummaryData` interface, `summaryData` to `RenderContext` |
| `src/config.ts` | Added `showSummary`, `summaryInterval` config options |
| `src/index.ts` | Wired `getSummary` into the main pipeline |
| `src/render/index.ts` | Appends summary lines after activity lines |
| `src/transcript.ts` | Added `extractRecentMessages()` for summary text extraction |
| `package.json` | Added `@anthropic-ai/claude-agent-sdk` as optional peer dep |

## Test plan

- [x] `npm run build` succeeds
- [x] Manually tested with `showSummary: true` and `summaryInterval: 1` across multiple terminals
- [x] Verified per-session cache isolation (no cross-terminal interference)
- [x] Verified graceful degradation when Agent SDK is not installed (summary simply doesn't appear)
- [x] Verified no render blocking (summary appears on next render after worker completes)


jarrodwatts:
thanks, appreciate the contribution, but just merged #155 which i think is a more simple solution to this

PR #131, opened 2026-02-20
feat(usage-api): add HTTP proxy support with NO_PROXY bypass

Hzfengsy:
## Summary

- Route usage API requests through an HTTPS proxy when standard proxy env vars are set (`HTTPS_PROXY`, `https_proxy`, `HTTP_PROXY`, `http_proxy`)
- Respect `NO_PROXY`/`no_proxy` to bypass proxy for specific hostnames (supports wildcard `*`, exact match, and domain suffix matching with/without leading dot)
- Support proxy authentication via URL credentials (`https://user:pass@proxy:8443`)

## Test plan

- [x] All 174 existing + new tests pass (`npm test`)
- [x] Added 9 tests for `isNoProxy` covering: no env, exact match, suffix with/without dot, wildcard, comma-separated list, case-insensitivity, lowercase env var, empty entries
- [x] Added 7 tests for `getProxyUrl` covering: no env, HTTPS_PROXY, preference order, HTTP_PROXY fallback, invalid URL, NO_PROXY interaction
- [x] Only `src/` and `tests/` modified — no `dist/` changes per CONTRIBUTING.md


liudongmiao:
https://nodejs.org/en/learn/http/enterprise-network-configuration#proxy-configuration

Newer nodejs support proxy in env. And, `https-proxy-agent` should be used instead a custom implemented one.

jarrodwatts:
Thanks for adding proxy and NO_PROXY handling. The same issue path is now merged via #164 in a narrower maintainer patch set, so I’m closing this as superseded.

PR #133, opened 2026-02-21
feat: add CLAUDE_CONFIG_DIR env var for custom config paths

ciberponk:
## Summary

- Add `CLAUDE_CONFIG_DIR` environment variable support so users can override the default `~/.claude` config directory
- On Windows, `os.homedir()` may not resolve to the expected config location; this env var provides a reliable override
- Changes affect three config-path resolution sites: `getConfigPath()`, `countConfigs()`, and `readFileCredentials()`

## Motivation

On Windows (and other non-standard setups), Claude Code may store its configuration in a directory that doesn't match `os.homedir() + '/.claude'`. For example, `os.homedir()` on Windows ignores the `HOME` environment variable and uses `USERPROFILE` instead. Setting `CLAUDE_CONFIG_DIR` allows the HUD plugin to find credentials, settings, and plugin config regardless of the OS-level home directory resolution.

## Changes

| File | Change |
|------|--------|
| `src/config.ts` | `getConfigPath()` checks `CLAUDE_CONFIG_DIR` before falling back to `~/.claude` |
| `src/config-reader.ts` | `countConfigs()` uses `CLAUDE_CONFIG_DIR` for locating settings, CLAUDE.md, and rules |
| `src/usage-api.ts` | `readFileCredentials()` tries `CLAUDE_CONFIG_DIR` first when looking for `.credentials.json` |

## Test plan

- [ ] Set `CLAUDE_CONFIG_DIR` to a custom path and verify HUD reads config from that path
- [ ] Unset `CLAUDE_CONFIG_DIR` and verify default `~/.claude` behavior is unchanged
- [ ] `npm run build` compiles without errors

### DIFF
--- a/src/config-reader.ts
+++ b/src/config-reader.ts
@@ -97,7 +97,11 @@ export async function countConfigs(cwd?: string): Promise<ConfigCounts> {
   let hooksCount = 0;
 
   const homeDir = os.homedir();
-  const claudeDir = path.join(homeDir, '.claude');
+  // Support CLAUDE_CONFIG_DIR env var for custom config locations (e.g. Windows)
+  const configDir = process.env.CLAUDE_CONFIG_DIR;
+  const claudeDir = configDir || path.join(homeDir, '.claude');
+  // If CLAUDE_CONFIG_DIR is set, derive the parent for .claude.json; otherwise use homeDir
+  const claudeJsonParent = configDir ? path.dirname(configDir) : homeDir;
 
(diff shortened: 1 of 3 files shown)


jarrodwatts:
Thank you for the contribution and for digging into this issue. Closing this PR as superseded by #160, which has now been merged with the consolidated CLAUDE_CONFIG_DIR and keychain handling changes.

PR #136, opened 2026-02-22
feat: add showProject config to hide project name

fredrikaverpil:
## Why

Some users prefer a more compact statusline without the project name, especially when the project context is obvious from the terminal/editor.

I for example, would like to have this:

```
[Opus 4.6 | Team] ████░░░░░░ 35% | ██░░░░░░░░ 20% (3h 14m / 5h) | ███░░░░░░░ 26% (96h 14m / 7d)
```

But as it is right now, I cannot hide the project name (folder name) that gets printed after the context usage.

## What

- Added `display.showProject` boolean config option (defaults to `true` for backwards compatibility)
- When set to `false`, hides the project path and its associated git info from both compact and expanded layouts
- Updated the configure skill to include the new option in turn on/off flows and element mapping
- Added test covering the new behavior

## Usage

Set in `~/.claude/plugins/claude-hud/config.json`:

```json
{
  "display": {
    "showProject": false
  }
}
```

## Test plan

- [x] All existing tests pass (159/159)
- [x] New test verifies project name is hidden when `showProject: false`
- [x] Backwards compatible — defaults to `true`, existing configs unaffected


jarrodwatts:
thank you @fredrikaverpil !!

PR #140, opened 2026-02-23
fix: support HTTPS_PROXY env var for usage API requests

liuhui998:
## Problem

Node.js's built-in `https` module does not automatically respect proxy environment variables (`HTTPS_PROXY`, `https_proxy`, `ALL_PROXY`, `all_proxy`). This causes the usage API request to fail with **403 Forbidden** for users whose traffic must route through an HTTP proxy — common in corporate networks and in regions where direct connections to `api.anthropic.com` are restricted .

The HUD then caches the failure and displays `Usage ⚠ (403)` indefinitely.

## Solution

Add `createProxyAgent()` in `usage-api.ts` that:

1. Reads standard proxy env vars (`HTTPS_PROXY` / `https_proxy` / `ALL_PROXY` / `all_proxy`)
2. Creates a custom `https.Agent` subclass that tunnels connections through the proxy using HTTP CONNECT
3. Passes the agent to `https.request()` — no-op when no proxy is configured

**Zero new dependencies** — implemented entirely with built-in Node.js modules (`net`, `tls`).

## Test plan

- [x] `npm run build` passes (TypeScript compiles cleanly)
- [x] `npm test` passes (158 pass, 1 skipped for Windows)
- [x] Manually verified: with `HTTPS_PROXY` set, the usage API returns 200 and correct utilization data; without `HTTPS_PROXY`, behavior is unchanged

### DIFF
--- a/src/usage-api.ts
+++ b/src/usage-api.ts
@@ -1,7 +1,10 @@
 import * as fs from 'fs';
 import * as path from 'path';
 import * as os from 'os';
+import * as http from 'http';
 import * as https from 'https';
+import * as net from 'net';
+import * as tls from 'tls';
 import { execFileSync } from 'child_process';
 import type { UsageData } from './types.js';
 import { createDebug } from './debug.js';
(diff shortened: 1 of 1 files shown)


liuhui998:
Hey @jarrodwatts! Sharing this PR directly in case the notification gets buried. This fixes usage API failures for users behind HTTP proxies — happy to adjust anything based on your feedback.

liudongmiao:
https://nodejs.org/en/learn/http/enterprise-network-configuration#proxy-configuration

Newer nodejs support proxy in env. And, `https-proxy-agent` should be used instead a custom implemented one.

Petersburg16:
I am really looking forward to this PR being merged into the official version; it is very important to me.

liudongmiao:
@Petersburg16 You can just upgrade node to v22.21.0 or v24.5.0+, and set environment `NODE_USE_ENV_PROXY` to `1`.

jarrodwatts:
Appreciate the proxy support patch. I merged a hardened proxy + NO_PROXY + timeout + OAuth fallback fix set in #164, so this PR is being closed as superseded.

PR #141, opened 2026-02-23
fix: prevent double-counting CLAUDE.md when cwd is home directory

liqiorg:
## Problem

When running Claude Code from the home directory (`~`), the HUD displays `2 CLAUDE.md` instead of `1 CLAUDE.md`.

This happens because `countConfigs()` in `config-reader.ts` checks both:
- **User scope**: `~/.claude/CLAUDE.md`
- **Project scope**: `{cwd}/.claude/CLAUDE.md`

When `cwd` is the home directory, both paths resolve to the same file, causing a double count.

## Fix

Skip the project scope `.claude/CLAUDE.md` and `.claude/CLAUDE.local.md` checks when `cwd === os.homedir()`, since those files are already counted in the user scope.

## Testing

- Build passes (`npm run build`)
- When cwd is home: shows `1 CLAUDE.md` (was `2 CLAUDE.md`)
- When cwd is a project dir: behavior unchanged

### DIFF
--- a/src/config-reader.ts
+++ b/src/config-reader.ts
@@ -187,17 +187,23 @@ export async function countConfigs(cwd?: string): Promise<ConfigCounts> {
     }
 
     // {cwd}/.claude/rules/*.md (recursive)
-    rulesCount += countRulesInDir(path.join(cwd, '.claude', 'rules'));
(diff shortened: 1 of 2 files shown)


jarrodwatts:
thanks @liqiorg !

PR #143, opened 2026-02-23
fix: use curl for usage API to bypass Cloudflare TLS blocking

WayneWang24:
## Summary

- **Use curl instead of Node.js `https` module for the usage API call** — Node.js gets blocked by Cloudflare's TLS fingerprinting (JA3/JA4) with a 403 "Request not allowed". curl has a trusted TLS fingerprint and works reliably.
- **Support `CLAUDE_CONFIG_DIR` for multi-account setups** — `readFileCredentials` now checks `CLAUDE_CONFIG_DIR` first, then falls back to `~/.claude/`. This fixes usage display when running Claude Code with alternate config directories.
- **Automatic proxy support** — curl natively respects `https_proxy`/`HTTPS_PROXY` environment variables, so proxy users get support for free without any manual tunnel handling.

## Problem

The usage API (`/api/oauth/usage`) sits behind Cloudflare. Cloudflare's bot detection uses TLS fingerprinting (JA3/JA4) to distinguish between clients. Node.js `https` module and built-in `fetch()` both have TLS fingerprints that Cloudflare blocks with a 403 "Request not allowed" response. This affects users in regions where requests go through certain Cloudflare PoPs.

## Solution

Replace `https.request()` with `execFileSync('/usr/bin/curl', ...)` — the same pattern already used for macOS Keychain access. Benefits:

1. curl has a trusted TLS fingerprint that passes Cloudflare's checks
2. curl natively handles `https_proxy`/`HTTPS_PROXY` env vars
3. Simpler code (no manual Promise/stream handling)
4. Security: uses absolute path `/usr/bin/curl` to prevent PATH hijacking

## Test plan

- [x] All 158 existing tests pass (0 failures)
- [x] Added `CLAUDE_CONFIG_DIR` isolation to test setup/teardown to prevent environment pollution
- [x] Verified curl-based fetch returns correct usage data (fiveHour, sevenDay, resetAt)
- [x] Verified proxy support works automatically through env vars
- [x] Tested with Team plan account behind proxy — confirmed 200 response where Node.js got 403


jarrodwatts:
Thanks for investigating Cloudflare/proxy failures. We’re not adopting a curl subprocess path for this flow due to safety/portability concerns. I merged a Node-based hardened usage path in #164, so I’m closing this as superseded.

PR #144, opened 2026-02-23
fix: handle object layout values in migrateConfig()

Rektoooooo:
## Problem

`migrateConfig()` checks `'layout' in userConfig` to detect the deprecated
legacy format, but doesn't verify `typeof layout === 'string'`. When a
third-party config editor writes `layout` as an object:

```json
{ "layout": { "lineLayout": "expanded", "showSeparators": true, "pathLevels": 2 } }
```

The migration incorrectly matches, overwrites with `lineLayout: "compact"`,
and deletes the user's settings.

## Fix

- Add `typeof` guard: only run legacy string migration when `layout` is a string
- When `layout` is an object, extract nested `lineLayout`, `showSeparators`,
  `pathLevels` to top level (graceful migration)
- Widen `LegacyConfig` type to accept `Record<string, unknown>`
- Export `mergeConfig` for direct testing of migration paths
- Add test cases for all migration scenarios

## Testing

- Legacy `layout: "default"` → migrates to compact (unchanged behavior)
- Legacy `layout: "separators"` → migrates to compact + separators (unchanged)
- Object `layout: { lineLayout: "expanded" }` → extracts to top level (new)
- Empty object `layout: {}` → falls back to defaults, no crash (new)
- No `layout` key → no migration (unchanged)
- Both `layout` and `lineLayout` present → `layout` ignored (unchanged)

All 164 tests pass (1 skipped — Windows-only path test).

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -89,19 +89,28 @@ function validateContextValue(value: unknown): value is ContextValueMode {
 }
 
 interface LegacyConfig {
-  layout?: 'default' | 'separators';
+  layout?: 'default' | 'separators' | Record<string, unknown>;
 }
 
 function migrateConfig(userConfig: Partial<HudConfig> & LegacyConfig): Partial<HudConfig> {
   const migrated = { ...userConfig } as Partial<HudConfig> & LegacyConfig;
 
   if ('layout' in userConfig && !('lineLayout' in userConfig)) {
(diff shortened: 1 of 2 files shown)


jarrodwatts:
Thanks @Rektoooooo !

PR #152, opened 2026-02-28
fix: support Claude Code 2.1+ profile-specific Keychain service names

9bow:
## Problem

Claude Code 2.1+ changed the macOS Keychain service name from the static `'Claude Code-credentials'` to profile-specific names with a hash suffix:

```
Claude Code-credentials-{hash}
```

This caused `readKeychainCredentials()` to silently fail on every render, resulting in `getUsage()` returning `null` and **the HUD showing no usage data** for all Claude Code 2.1+ users.

Users who re-login still experience the issue because the token refresh preserves the profile-specific service name.

## Root Cause

```typescript
// Before: hardcoded legacy service name
const keychainData = execFileSync(
  '/usr/bin/security',
  ['find-generic-password', '-s', 'Claude Code-credentials', '-w'],
  ...
);
```

## Fix

1. **Enumerate all matching Keychain entries** using `security dump-keychain`, matching the pattern `Claude Code-credentials(-[a-f0-9]+)?`
2. **Try each entry** and return the first valid credentials
3. **Profile-aware selection**: when `CLAUDE_CONFIG_DIR` is set (e.g. `~/.claude-max`), prefer the entry whose `subscriptionType` matches the profile name — useful for users running multiple Claude Code accounts simultaneously
4. **Backward compatible**: falls back to `'Claude Code-credentials'` (legacy) when no profile-specific entries are found

## Test

Verified on macOS with Claude Code 2.1.63 running two profiles (`max` and `pro`):

```bash
CLAUDE_CONFIG_DIR=~/.claude-max bun --eval "
import { getUsage } from './src/usage-api.ts';
console.log(JSON.stringify(await getUsage(), null, 2));
"
# Output:
# {
#   "planName": "Max",
#   "fiveHour": 59,
#   "sevenDay": 16,
#   ...
# }
```

### DIFF
--- a/src/usage-api.ts
+++ b/src/usage-api.ts
@@ -3,6 +3,7 @@ import * as path from 'path';
 import * as os from 'os';
 import * as https from 'https';
 import { execFileSync } from 'child_process';
+import { createHash } from 'crypto';
 import type { UsageData } from './types.js';
 import { createDebug } from './debug.js';
 
@@ -40,7 +41,7 @@ interface UsageApiResult {
 // File-based cache (HUD runs as new process each render, so in-memory cache won't persist)
 const CACHE_TTL_MS = 60_000; // 60 seconds
 const CACHE_FAILURE_TTL_MS = 15_000; // 15 seconds for failed requests
-const KEYCHAIN_TIMEOUT_MS = 5000;
+const KEYCHAIN_TIMEOUT_MS = 3000;
 const KEYCHAIN_BACKOFF_MS = 60_000; // Backoff on keychain failures to avoid re-prompting
 
 interface CacheFile {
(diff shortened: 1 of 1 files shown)


9bow:
## Updated: Direct hash lookup + API timeout fix

This PR has been updated with additional fixes for multi-profile setups.

### Changes (v2)

**1. Replace `security dump-keychain` with direct hash computation**

The previous approach used `security dump-keychain` to enumerate all keychain entries, which:
- Dumps the **entire macOS keychain** (slow, 5s+ on large keychains)
- Falls back to `Claude Code-credentials` (no hash) on timeout — which **doesn't exist** when using `CLAUDE_CONFIG_DIR`

Now computes the exact service name directly:
```typescript
function getKeychainServiceName(): string {
  const configDir = process.env.CLAUDE_CONFIG_DIR;
  if (configDir) {
    const hash = createHash('sha256').update(configDir).digest('hex').slice(0, 8);
    return `Claude Code-credentials-${hash}`;
  }
  return 'Claude Code-credentials';
}
```

This matches how Claude Code itself stores credentials (one keychain entry per `CLAUDE_CONFIG_DIR` hash).

**2. Increase API timeout: 5s → 15s**

`api.anthropic.com/api/oauth/usage` can take 9+ seconds to respond. The previous 5s timeout caused consistent `Usage ⚠ (timeout)` in the HUD.

**3. Reduce keychain timeout: 5s → 3s**

Direct `find-generic-password` lookup is much faster than `dump-keychain`, so 3s is more than sufficient.

**4. Remove fragile profile matching**

The previous code matched profiles by checking if `subscriptionType.toLowerCase().includes(profileHint)` (e.g., `.claude-max` directory → "max" hint → matches "max" subscription type). This was coincidental and unreliable. The hash-based service name already uniquely identifies the correct profile.

### Summary

| Metric | Before | After |
|--------|--------|-------|
| Keychain lookup | `dump-keychain` + regex (5s+) | Direct hash lookup (<100ms) |
| API timeout | 5,000ms | 15,000ms |
| Keychain timeout | 5,000ms | 3,000ms |
| Profile matching | subscriptionType name match | Hash-based (exact) |

### Testing

Verified with `CLAUDE_CONFIG_DIR=/Users/[USER]/.claude-max`:
- `[Opus | Max]` — plan name correctly displayed
- `Usage 9% (3h 46m / 5h) | 25% (86h 46m / 7d)` — usage data fetched successfully
- Cached response: 0.037s

jarrodwatts:
Thank you for the contribution and for digging into this issue. Closing this PR as superseded by #160, which has now been merged with the consolidated CLAUDE_CONFIG_DIR and keychain handling changes.

PR #155, opened 2026-03-01
feat: show session name in statusline

myaiexp:
## Summary

- Reads the session slug (auto-generated) and custom title (set via `/rename`) from the transcript JSONL
- Displays it in dim text after the project/git info on both expanded (project line) and compact (session line) layouts
- Custom title takes priority over auto-generated slug when both exist

## Motivation

Makes it easy to see at a glance whether a session has been named and what it's named, without running any command. Particularly useful as a prompt to run `/rename` when you see the auto-generated slug still showing.

## How it works

Claude Code writes a `slug` field on every transcript entry and a `custom-title` typed entry when `/rename` is used. The parser now tracks both as it scans the JSONL, preferring the custom title if set.

### DIFF
--- a/tests/core.test.js
+++ b/tests/core.test.js
@@ -186,6 +186,43 @@ test('parseTranscript aggregates tools, agents, and todos', async () => {
   assert.equal(result.sessionStart?.toISOString(), '2024-01-01T00:00:00.000Z');
 });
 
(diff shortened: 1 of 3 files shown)


bill-kolokithas:
Hello. I would like config option to remove that as it provides no value to me.

PR #164, opened 2026-03-05
fix: address Windows setup, Bedrock labels, and usage API reliability

### LINKED ISSUE #121: /claude-hud:setup generates broken command on Windows (win32)

### LINKED ISSUE #137: Bedrock startup sometimes shows raw model id (eu.anthropic...) instead of model name
## Summary
When Claude Code starts a session on AWS Bedrock and `model.display_name` is not yet present in stdin, Claude HUD shows the raw Bedrock model ID (for example `eu.anthropic.claude-opus-4-5-20251101-v1:0`) in the statusline instead of a readable model name.

## Steps to Reproduce
1. Install current `main` version of Claude HUD.
2. Run Claude Code with a Bedrock-backed model where startup frames may include only `model.id`.
3. Or reproduce directly with stdin:

```bash
echo '{"model":{"id":"eu.anthropic.claude-opus-4-5-20251101-v1:0"},"cwd":"/tmp","context_window":{"current_usage":{"input_tokens":1000},"context_window_size":200000}}' | node dist/index.js
```

4. Observe the rendered first line.

## Expected Behavior
Model label should remain human-readable (for example `Opus`, `Sonnet`, or an intentionally normalized Bedrock model label), not the raw provider-qualified model ID.

## Actual Behavior
HUD shows:

```text
[eu.anthropic.claude-opus-4-5-20251101-v1:0 | Bedrock]
```

This is noisy and appears intermittently during startup depending on whether `display_name` is present yet.

## Environment

- OS: macOS
- Node/Bun version: Node.js 22.x
- Claude Code version: 2.x

## Logs or Screenshots
Observed behavior maps to current fallback logic:

- `src/stdin.ts`:
  - `getModelName()` currently returns `stdin.model?.display_name ?? stdin.model?.id ?? 'Unknown'`
- Bedrock detection and provider label are already correct (`Bedrock`), so the issue is specifically model display fallback when `display_name` is missing.

### LINKED ISSUE #146: Usage API timeout too short for WSL2 (5s hardcoded, response takes ~5.3s)
## Summary

The `fetchUsageApi` timeout is hardcoded to 5000ms (`src/usage-api.ts:399`). On WSL2, Node.js `https.request` to `api.anthropic.com/api/oauth/usage` consistently takes ~5.2–5.3s due to WSL2 network stack latency (DNS resolution + NAT traversal through the Windows host). The request completes successfully with a 10s timeout — it's not a network failure, just marginal latency exceeding the tight default.

## Steps to Reproduce

1. Run Claude Code inside WSL2 (Ubuntu)
2. Enable `showUsage: true` in HUD config
3. Observe `Usage: ⚠ (timeout)` in the HUD status line
4. Check `~/.claude/plugins/claude-hud/.usage-cache.json` — shows `"apiError": "timeout"`

## Expected Behavior

Usage bar displays utilization percentages, e.g. `65% (31m / 5h) | 6% (67h / 7d)`

## Actual Behavior

Usage permanently shows `⚠ (timeout)`. The failure caches for 15s (`CACHE_FAILURE_TTL_MS`), then retries and times out again in a loop. Usage data is never displayed.

## Environment

- OS: WSL2 — Ubuntu 24.04 on Windows 11 (kernel 6.6.87.2-microsoft-standard-WSL2)
- Node/Bun version: Node v24.11.0
- Claude Code version: v2.1.51
- claude-hud version: 0.0.7

## Logs or Screenshots

Diagnostic test — Node.js `https.request` to the same endpoint with a 10s timeout:

```
Status: 200 Time: 5279 ms
Body: {"five_hour":{"utilization":65.0,"resets_at":"2026-02-24T07:00:00..."}, ...}
```

The request succeeds at 5279ms — just 279ms over the 5000ms limit. Python `urllib` with 10s timeout succeeds consistently. Bumping the hardcoded timeout from 5000 to 10000ms in `dist/usage-api.js` fixes the issue completely.

**Cached failure result:**
```json
{
  "data": {
    "planName": "Max",
    "fiveHour": null,
    "sevenDay": null,
    "apiUnavailable": true,
    "apiError": "timeout"
  }
}
```

## Suggested Fix

Either:
1. **Bump the default timeout to 10s** — since results cache for 60s, the extra 5s on first fetch has negligible UX impact
2. **Make the timeout configurable** via `config.json` (e.g., `usageApiTimeoutMs`) so users in high-latency environments can tune it

Option 1 is simpler and sufficient. WSL2 is a common dev environment and the current 5s default is too tight for it.

## Related

- #143 — also about usage API failures, but different root cause (Cloudflare TLS fingerprint blocking vs. WSL2 network latency)

### LINKED ISSUE #148: Windows: setup generates broken statusLine command when Claude Code uses Git Bash as shell
## Summary
When running Claude Code on Windows with Git Bash as the shell, the '/claude-hud:setup' skill generates a PowerShell-based 'statusLine' command that silently breaks at runtime. Instead of the HUD, a long raw JSON string appears below the chat input.

## Steps to Reproduce
1. Install Claude Code on Windows with Git Bash as the configured shell
2.  Install claude-hud via /plugin install claude-hud
3. Run /claude-hud:setup
4. Observe the statusLine area, a raw JSON appears instead of the HUD

## Expected Behavior
The HUD renders correctly below the chat input.

## Actual Behavior
A large raw JSON blob appears, starting with something like:{"session_id":"...","transcript_path":"C:\Users\...","cwd":"E:\..."}

## Environment
- OS: Windows 11
- Claude Code shell: Git Bash ('bash')
- Platform value seen by Claude Code: 'win32'
- claude-hud version: 0.0.7
- Claude Code v2.1.56

## Root Cause

  The setup skill detects `Platform: win32` and generates a PowerShell command:

  powershell -Command "& {$p=(Get-ChildItem $env:USERPROFILE.claude\plugins\cache\claude-hud\claude-hud | Sort-Object LastWriteTime -Descending | Select-Object -First 1).FullName; & 'C:\nvm4w\nodejs\node.exe'
  (Join-Path $p 'dist\index.js')}"

  When Claude Code invokes this `statusLine` command through **bash** (Git Bash), bash
  processes the string **before** passing it to PowerShell and expands:

  - `$p` → empty string (undefined bash variable)
  - `$env` → empty string, leaving `:USERPROFILE\...` as a literal path fragment

  The resulting command arrives at PowerShell completely corrupted, fails silently,
  and Claude Code falls back to displaying the raw context JSON it tried to pass as stdin.

## Fix Applied (Workaround)

  Replacing the PowerShell command in `~/.claude/settings.json` with a bash-compatible
  equivalent fixes the issue:

  ```json
  {
    "statusLine": {
      "type": "command",
      "command": "bash -c '\"/c/nvm4w/nodejs/node\" \"$(ls -td ~/.claude/plugins/cache/claude-hud/claude-hud/*/ 2>/dev/null | head -1)dist/index.js\"'"
    }
  }
```
## Suggested Fix for the Setup Skill

  The setup skill already has access to the Shell: value in Claude Code's environment
  context. On win32, it should branch on the shell:

  - Shell is bash (Git Bash) → generate a bash-compatible command (same format as macOS/Linux but with Windows node path)
  - Shell is cmd or powershell → keep the current PowerShell command

## Additional Notes

  This likely became a widespread issue if Claude Code recently started defaulting to
  Git Bash on Windows, as the PowerShell command would have worked previously when
  commands were dispatched through cmd.exe.

### LINKED ISSUE #161: Usage stats and plan label disappear after Claude Code 2.1.63 token refresh
## Bug

After updating to **Claude Code 2.1.63**, the plan label (`| Max`) and usage stats (`5h: X% | 7d: X%`) disappear from the HUD once the usage cache expires (60s TTL).

## Root Cause

Claude Code 2.1.63 no longer stores `subscriptionType` in the keychain entry when refreshing the OAuth token — the field is explicitly `null`. You can verify:

```bash
security find-generic-password -s "Claude Code-credentials" -w | python3 -c \
  "import json,sys; d=json.load(sys.stdin); print(d['claudeAiOauth'].get('subscriptionType'))"
# Output: None
```

The 0.0.8 change (#158) made `getPlanName` treat an empty `subscriptionType` as an API user, causing `getUsage()` to return `null` before ever calling the API:

```typescript
if (!subscriptionType || lower.includes('api')) return null; // breaks OAuth users with null subscriptionType
```

An empty `subscriptionType` doesn't necessarily mean an API user — in Claude Code 2.1.63, OAuth subscribers also have this field as `null`.

## Environment

- macOS 25.3.0, Claude Code 2.1.63, claude-hud 0.0.8, Max plan

### LINKED ISSUE #162: Usage warning(403) when network proxy is on
The current version shows a Usage warning (403) when a network proxy is enabled. I used Claude to fix this issue and generated a corresponding .md file. Please consider addressing this problem in main branch.

[claude-hud-proxy-fix.md](https://github.com/user-attachments/files/25735214/claude-hud-proxy-fix.md)

jarrodwatts:
## Summary
- Fix Windows setup instructions to branch correctly on `Platform: win32` + `Shell: bash` vs PowerShell/cmd.
- Normalize Bedrock model IDs when `model.display_name` is missing so startup labels stay human-readable.
- Harden usage fetch behavior:
  - proxy env support: `HTTPS_PROXY`/`ALL_PROXY`/`HTTP_PROXY` with `NO_PROXY` bypass
  - OAuth token-refresh fallback when keychain token exists but `subscriptionType` is missing
  - safer default timeout (15s) with env override via `CLAUDE_HUD_USAGE_TIMEOUT_MS`
- Document the timeout override in `README.md`.

## Safety and Scope
- Narrow bug-fix scope only (no feature expansion).
- No new runtime dependencies.
- Source/docs/tests changes only.

## Validation
- `npm run build`
- `npm test` (212 pass, 0 fail, 1 skip)

## Closes
- Closes #121
- Closes #148
- Closes #137
- Closes #146
- Closes #161
- Closes #162

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,8 @@ To disable, set `display.showUsage` to `false`.
 - Check `display.showUsage` is not set to `false` in config
 - API users see no usage display (they have pay-per-token, not rate limits)
 - AWS Bedrock models display `Bedrock` and hide usage limits (usage is managed in AWS)
+- If you are behind a proxy, set `HTTPS_PROXY` (or `HTTP_PROXY`/`ALL_PROXY`) and optional `NO_PROXY`
+- For high-latency environments, increase usage API timeout with `CLAUDE_HUD_USAGE_TIMEOUT_MS` (milliseconds)
 
 ### Example Configuration
 
--- a/src/usage-api.ts
+++ b/src/usage-api.ts
@@ -1,6 +1,8 @@
 import * as fs from 'fs';
 import * as path from 'path';
 import * as os from 'os';
+import * as net from 'net';
+import * as tls from 'tls';
 import * as https from 'https';
 import { execFileSync } from 'child_process';
 import { createHash } from 'crypto';
@@ -45,6 +47,7 @@ const CACHE_TTL_MS = 60_000; // 60 seconds
 const CACHE_FAILURE_TTL_MS = 15_000; // 15 seconds for failed requests
 const KEYCHAIN_TIMEOUT_MS = 3000;
 const KEYCHAIN_BACKOFF_MS = 60_000; // Backoff on keychain failures to avoid re-prompting
+const USAGE_API_TIMEOUT_MS_DEFAULT = 15_000;
 
 interface CacheFile {
   data: UsageData;
@@ -387,6 +390,30 @@ function readFileCredentials(homeDir: string, now: number): { accessToken: strin
   }
 }
 
+function readFileSubscriptionType(homeDir: string): string | null {
+  const credentialsPath = path.join(getClaudeConfigDir(homeDir), '.credentials.json');
+
+  if (!fs.existsSync(credentialsPath)) {
+    return null;
+  }
+
+  try {
+    const content = fs.readFileSync(credentialsPath, 'utf8');
+    const data: CredentialsFile = JSON.parse(content);
+    const subscriptionType = data.claudeAiOauth?.subscriptionType;
+    const normalizedSubscriptionType = typeof subscriptionType === 'string'
+      ? subscriptionType.trim()
+      : '';
+    if (!normalizedSubscriptionType) {
+      return null;
+    }
+    return normalizedSubscriptionType;
+  } catch (error) {
+    debug('Failed to read file subscriptionType:', error);
+    return null;
+  }
+}
+
 /**
  * Parse and validate credentials data from either Keychain or file.
  */
@@ -429,12 +456,12 @@ function readCredentials(
       return keychainCreds;
     }
     // Keychain has token but no subscriptionType - try to supplement from file
-    const fileCreds = readFileCredentials(homeDir, now);
-    if (fileCreds?.subscriptionType) {
+    const fileSubscriptionType = readFileSubscriptionType(homeDir);
+    if (fileSubscriptionType) {
       debug('Using keychain token with file subscriptionType');
       return {
         accessToken: keychainCreds.accessToken,
-        subscriptionType: fileCreds.subscriptionType,
+        subscriptionType: fileSubscriptionType,
       };
     }
     // No subscriptionType available - use keychain token anyway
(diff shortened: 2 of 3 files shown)


liudongmiao:
@jarrodwatts Could you check whether  `NODE_USE_ENV_PROXY` is set, so we can just use the proxy in nodejs? https://nodejs.org/en/learn/http/enterprise-network-configuration#proxy-configuration

PR #167, opened 2026-03-05
fix(usage-api): proxy CONNECT tunnel broken by premature socket return

xiangboit:
## Problem

The proxy CONNECT tunnel added in #164 consistently fails with `400 Bad Request` from proxies like Clash, causing the Usage display to show a persistent `(403)` warning.

### Root Cause

In `createProxyTunnelAgent`, `createConnection` returns the raw proxy socket at [line 643](https://github.com/jarrodwatts/claude-hud/blob/b57aa1e/src/usage-api.ts#L643):

```ts
return proxySocket;
```

Node.js `_http_agent.js` → `createSocket()` does:

```js
const oncreate = once((err, s) => { /* register socket, start HTTP write */ });

const newSocket = self.createConnection(options, oncreate);
if (newSocket && !newSocket[kWaitForProxyTunnel])
    oncreate(null, newSocket);
```

Because `proxySocket` is truthy and `kWaitForProxyTunnel` is **not set** (it's a private internal symbol only set by Node.js's own HTTPS proxy handling in `lib/https.js`), `oncreate(null, proxySocket)` fires immediately with the **raw TCP socket** — before the CONNECT handshake even begins.

Since `oncreate` is wrapped with `once()`, when the CONNECT tunnel later completes and the callback calls `oncreate(null, tlsSocket)` with the proper TLS socket, **it is silently dropped**.

The result: Node.js writes the HTTP request directly to the proxy's plain TCP connection. The proxy sees an invalid request and responds with `400 Bad Request`.

### Debug trace

```
[claude-hud:usage] Using proxy for usage API: http://127.0.0.1:7890
[claude-hud:usage] API request error: Error: Proxy CONNECT rejected: HTTP/1.1 400 Bad Request
```

## Fix

Return `undefined` from `createConnection` instead of `proxySocket`. When the return value is falsy, Node.js skips the immediate `oncreate` call and waits for the async callback — which fires only after the CONNECT tunnel + TLS handshake succeeds.

```diff
-      return proxySocket;
+      // Must not return the socket here. In Node.js _http_agent.js, createSocket()
+      // calls: `if (newSocket && !newSocket[kWaitForProxyTunnel]) oncreate(null, newSocket)`
+      // — returning a truthy value causes oncreate to fire immediately with the raw
+      // proxy socket (before CONNECT handshake), and once() then silently drops the
+      // later callback with the correct TLS socket. Return undefined so Node.js waits
+      // for the async callback after CONNECT + TLS succeeds.
+      return undefined;
```

One-line change in `src/usage-api.ts`.

## Reproduction

Minimal repro script (tested on Node.js v25.4.0, macOS, Clash proxy at `127.0.0.1:7890`):

```js
// ❌ Buggy — returns raw socket, Node.js uses it immediately
createConnection(options, callback) {
  const proxySocket = net.connect(proxyPort, proxyHost);
  proxySocket.once('connect', () => { /* CONNECT handshake → callback(null, tlsSocket) */ });
  return proxySocket;  // oncreate(null, proxySocket) fires now, once() blocks the later TLS callback
}

// ✅ Fixed — Node.js waits for the async callback with the TLS socket
createConnection(options, callback) {
  const proxySocket = net.connect(proxyPort, proxyHost);
  proxySocket.once('connect', () => { /* CONNECT handshake → callback(null, tlsSocket) */ });
  return undefined;
}
```

**Before fix**: `Proxy CONNECT rejected: HTTP/1.1 400 Bad Request` → Usage shows `⚠ (403)`
**After fix**: `HTTP/1.1 200 Connection established` → Usage displays normally (e.g. `5h 29% | 7d 29%`)

## Validation

- `npm run build` ✅
- `npm test` — 213 pass, 0 fail, 1 skip ✅
- Manual verification with Clash proxy on macOS ✅
- Only `src/` modified — no `dist/` changes per CONTRIBUTING.md

## Related

- #164 (introduced the proxy tunnel agent)
- #146, #161, #162 (proxy-related usage API issues)

### DIFF
--- a/tests/usage-api.test.js
+++ b/tests/usage-api.test.js
@@ -15,6 +15,7 @@ import { tmpdir } from 'node:os';
 import path from 'node:path';
 import { createHash } from 'node:crypto';
 import { existsSync } from 'node:fs';
+import { createServer } from 'node:net';
 
 let tempHome = null;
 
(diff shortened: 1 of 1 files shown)


jarrodwatts:
thanks @xiangboit

PR #168, opened 2026-03-05
fix(usage-api): change User-Agent to avoid Anthropic 429 rate limiting

xiangboit:
## Problem

The usage API consistently returns **429 Too Many Requests** for many users, causing the Usage section to show `⚠` or disappear entirely.

The root cause is the `User-Agent: claude-hud/1.0` header. Anthropic applies **stricter rate limits to non-official User-Agent strings**.

## Evidence

Tested with the same OAuth token, same IP, same time — only the User-Agent differs:

```
$ curl -s -o /dev/null -w "%{http_code}" \
  -H "User-Agent: claude-hud/1.0" \
  -H "Authorization: Bearer $TOKEN" \
  -H "anthropic-beta: oauth-2025-04-20" \
  https://api.anthropic.com/api/oauth/usage
429

$ curl -s -o /dev/null -w "%{http_code}" \
  -H "User-Agent: claude-code/2.1" \
  -H "Authorization: Bearer $TOKEN" \
  -H "anthropic-beta: oauth-2025-04-20" \
  https://api.anthropic.com/api/oauth/usage
200
```

This is reproducible across different accounts and networks.

## Fix

Change the User-Agent from `claude-hud/1.0` to `claude-code/2.1`.

This is a one-line change in `src/usage-api.ts`.

## Discussion

This is intentionally submitted as a minimal fix to unblock users. There are a few directions the maintainer might prefer:

1. **Accept as-is** — pragmatic fix, matches what Claude Code itself sends
2. **Use a different UA** — e.g. `claude-code/2.1 (claude-hud/0.0.9)` to retain identification while keeping the official prefix
3. **Report upstream** — file an issue with Anthropic about the UA-based rate limiting so plugins can use their own identity

Happy to adjust the UA string based on your preference.

## Validation

- `npm run build` ✅
- `npm test` — 213 pass, 0 fail, 1 skip ✅
- Manual verification: Usage displays correctly after change ✅
- Only `src/` modified per CONTRIBUTING.md ✅

## Related

- #164 (added proxy support for usage API)
- #167 (fixed proxy CONNECT tunnel)

### DIFF
--- a/src/usage-api.ts
+++ b/src/usage-api.ts
@@ -48,6 +48,7 @@ const CACHE_FAILURE_TTL_MS = 15_000; // 15 seconds for failed requests
 const KEYCHAIN_TIMEOUT_MS = 3000;
 const KEYCHAIN_BACKOFF_MS = 60_000; // Backoff on keychain failures to avoid re-prompting
 const USAGE_API_TIMEOUT_MS_DEFAULT = 15_000;
+export const USAGE_API_USER_AGENT = 'claude-code/2.1';
 
 interface CacheFile {
   data: UsageData;
@@ -662,7 +663,7 @@ function fetchUsageApi(accessToken: string): Promise<UsageApiResult> {
       headers: {
         'Authorization': `Bearer ${accessToken}`,
         'anthropic-beta': 'oauth-2025-04-20',
-        'User-Agent': 'claude-code/2.1',
+        'User-Agent': USAGE_API_USER_AGENT,
       },
       timeout: timeoutMs,
       agent: proxyUrl ? createProxyTunnelAgent(proxyUrl) : undefined,
--- a/tests/usage-api.test.js
+++ b/tests/usage-api.test.js
@@ -9,6 +9,7 @@ import {
   getUsageApiTimeoutMs,
   isNoProxy,
   getProxyUrl,
+  USAGE_API_USER_AGENT,
 } from '../dist/usage-api.js';
 import { mkdtemp, rm, mkdir, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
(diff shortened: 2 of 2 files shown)


joncchen:
**Supporting this PR.** We were seeing frequent 429 timeouts from the usage API throughout today's sessions, causing the Usage section to show `⚠` or go blank. After applying this one-line UA change locally, the API returns 200 consistently and the 429s stopped completely.

Minimal change, clear root cause, and the PR author's evidence (same token, same IP, different UA → different status code) matches exactly what we observed. +1 to merge.

Dbochman:
Confirming this fix worked for me, I was similarly seeing 429's for usage.

jarrodwatts (approved):
Maintainer follow-up pushed with a narrow regression check. Verified locally with npm test and green CI on the updated branch.

xiangboit:
## Follow-up: Corrected analysis — rate limiting is per-token, not per-UA

After more controlled testing, I found my original analysis was incomplete. The 429s are **not caused by the User-Agent string itself**.

### Test methodology

Sent rapid sequential requests with the same OAuth token, varying only the User-Agent:

```
Group A: claude-hud/1.0     → #1-4: 200, #5-10: 429
Group B: claude-code/2.1    → #1-10: 429 (rate limit already hit by Group A)
Group C: unique random UA   → #1-10: 429 (same)
```

After cooldown, all three UA strings return 200 equally. The limit is **per-token with a sliding window (~4-5 req/min)**, not per-UA.

The only hard rule observed: **empty User-Agent always gets 429**, regardless of rate limit state.

### Why the UA change appeared to fix it

When I changed the UA and restarted the terminal, enough cooldown time had naturally passed. It was correlation, not causation.

### Real root cause for multi-terminal users

Multiple terminals share the same token but each independently fetches when their local cache expires (~60s TTL). With 3+ terminals, cache expiration overlap easily exceeds the ~4-5 req/min threshold.

A potential improvement would be cross-process request deduplication (e.g. file-based fetch lock), so only one terminal fetches at a time while others serve stale cache. Happy to contribute if that direction is welcome.

Apologies for the incorrect initial analysis. The merged change is harmless but not the fix I originally claimed.

xiangboit:
To add some context: this isn't just a theoretical edge case. As a daily Claude Code user with multiple terminal splits, the Usage display breaks frequently throughout the day — showing `⚠` or going blank for extended periods. The 60s cache helps but doesn't fully solve it when 3-4 terminals expire around the same time.

Would love to see this stabilized, whether through a fetch lock, longer cache TTL, or another approach. Happy to help implement whatever direction you prefer.

Issue #172, opened 2026-03-06
HUD lines truncated/hidden in narrow terminal splits


xiangboit:
## Problem

When using terminal splits (e.g. tmux panes, VS Code split terminal, iTerm2 split), the HUD output gets truncated by the terminal. Lines like **MCP servers**, **Bash/Edit/Glob tool counts**, and **todo progress** are pushed off-screen and replaced by system prompts.

Only the first 1-2 lines (model info + context bar) remain visible. The rest of the HUD content is effectively hidden.

## Environment

- macOS + iTerm2 / VS Code integrated terminal
- Terminal width: ~80 cols (split pane)
- claude-hud v0.0.9

## Expected Behavior

HUD should adapt to narrow terminals by either:
- Condensing information into fewer lines
- Prioritizing the most important lines (e.g. usage, context) when space is limited
- Collapsing less critical sections (tools, todos, agents) at narrow widths

## Current Behavior

All lines are rendered at full width regardless of available terminal space. In narrow splits, the lower lines are pushed below the visible area and get overwritten by Claude Code's own output.

## Possible Approaches

1. **Responsive line collapsing** — detect terminal width and hide low-priority lines when below a threshold
2. **Compact mode** — a config option for a minimal single-line or two-line layout
3. **Priority-based rendering** — always show usage + context, conditionally show tools/todos/agents based on available rows

Happy to help implement if there's a preferred direction.

ashishtiwari03:
Hi! I’d like to take a look at this issue. My initial idea is to detect terminal width and render a compact HUD layout when the width is below a threshold. Please let me know if that approach aligns with the intended direction before I start working on a fix.

jarrodwatts:
Closing as already resolved on current main. The renderer now wraps output to terminal width and keeps activity lines visible, with regression coverage in tests/render-width.test.js. If you can still reproduce on the latest release, please reopen with the exact viewport/terminal behavior you see.

PR #174, opened 2026-03-06
fix(usage): prevent cross-process usage API stampedes

jarrodwatts:
## Problem

The merged fix in #168 changed the `User-Agent`, but that did not address the real issue. The underlying failure mode is a cross-process cache stampede:

- Claude HUD runs as a new process on each render
- usage data is cached on disk for 60s on success / 15s on failure
- when that cache expires, multiple terminals can miss it at the same time
- each process then makes its own `/api/oauth/usage` request with the same token
- Anthropic rate limits per token, so concurrent refreshes can produce `429` responses and make usage show `⚠` or disappear

## Root Cause Investigation

I reproduced the issue deterministically in tests rather than relying on anecdotal `User-Agent` behavior.

The important finding is:

- the problem is not the specific non-empty `User-Agent`
- the problem is uncoordinated refreshes after cache expiry across multiple HUD processes

## Fix

This PR changes the usage cache flow to coordinate refreshes across processes:

- add a file lock so only one process refreshes usage at a time
- followers reuse stale cached usage while the leader refreshes expired data
- cold-start followers wait briefly for the fresh cache instead of issuing duplicate requests
- clean up the lock in `clearCache()`
- restore an honest HUD `User-Agent` instead of impersonating `claude-code`

## Tests

Added regression coverage for the actual issue:

- concurrent cache misses only trigger one fetch
- expired cache remains visible while a refresh is in flight
- `User-Agent` remains non-empty and identifies `claude-hud`

## Validation

- `npm test` ✅ (231 pass, 0 fail, 1 skip)

## Related

- Supersedes the root-cause analysis from #168


xiangboit:
Thanks for implementing this properly — the atomic lock with `O_EXCL` and the cold-start polling are solid improvements over what I had in mind.

This plugin has been a great addition to my daily workflow. The usage display, context tracking, and tool activity monitoring save me a lot of guesswork. Appreciate the responsiveness on issues and PRs.

One related item when you get a chance: #172 — in narrow terminal splits (tmux panes, VS Code split), the HUD lines get pushed off-screen by Claude Code's own output. Would be great to have some form of responsive condensing for smaller terminals. Happy to help if you'd like a contribution there.

PR #185, opened 2026-03-09
fix(render): silence usage warning when API returns 403

dsrmyao:
## Problem

  When the HUD plugin calls `/api/oauth/usage`, it sometimes receives a
  403 response because the API restricts access to the Claude Code main
  process. This is expected behaviour, not a user-actionable error.

  The HUD was showing `Usage ⚠ (403)` persistently, misleading users
  into thinking something was broken.

  ## Solution

  When a 403 is received, the Usage line is hidden silently instead of
  showing a warning. The HUD will continue retrying on each cache
  expiry (15s on failure), so usage data will appear automatically once
  a successful response comes through.

  Other errors (network failure, timeout, etc.) still surface the `⚠`
  warning as before.


jarrodwatts:
Closing in favor of fixing the underlying usage-path issues rather than hiding  output. The current maintenance direction is to harden version selection and rate-limit handling instead of suppressing a status that can still represent a real configuration/runtime problem.

Issue #187, opened 2026-03-09
Usage query rate-limited (429)


tomholford:
Usage showing 429 (rate-limited) on occasion:

<img width="146" height="42" alt="Image" src="https://github.com/user-attachments/assets/8354b6f4-5a38-4484-9e4d-a6cabf8238b0" />

At the moment, have 2 active parallel sessions both using claude-hud.

Claude Code v2.1.71
claude-hud 0.0.9

YASoftwareDev:
Related: #188 — the 15s  makes this worse by retrying 4×/minute after a 429, turning a transient rate-limit into a persistent warning with no recovery path. Proposed fix there: bump the failure TTL to 5 minutes.

ngreenwood6:
bump

jarrodwatts:
Closed by merged PR #193, which rebases the 429 handling onto current main and adds cache/backoff hardening plus regression coverage.

PR #189, opened 2026-03-09
fix: increase failure cache TTL from 15s to 5min to prevent 429 retry storm

### LINKED ISSUE #188: Usage API: increase failure cache TTL to avoid 429 retry storm
## Problem

When the `/api/oauth/usage` endpoint returns a 429 (rate limited), the current `CACHE_FAILURE_TTL_MS` of **15 seconds** causes the HUD to retry the API 4 times per minute indefinitely. This creates a retry storm that keeps hitting the rate limit, making recovery impossible and showing a persistent `Usage ⚠ (429)` warning in the status bar.

## Root Cause

In `src/usage-api.ts`:

```ts
const CACHE_FAILURE_TTL_MS = 15_000; // 15 seconds for failed requests
```

After a 429, the HUD backs off for only 15s, then retries — which triggers another 429, then another 15s backoff, and so on.

## Suggested Fix

Increase the failure TTL to at least **5 minutes** (preferably with exponential backoff, but a fixed longer TTL already fixes the retry storm):

```ts
const CACHE_FAILURE_TTL_MS = 300_000; // 5 minutes for failed requests
```

While at it, `CACHE_TTL_MS = 60_000` (60s for success) could also be bumped to 5 minutes — usage data doesn't change meaningfully within seconds, and the 5-hour rolling window only shifts over minutes.

## Impact

- Persistent `Usage ⚠ (429)` warning that never clears
- Unnecessary API traffic that prolongs the rate-limit period
- No way to recover without manually deleting `~/.claude/plugins/claude-hud/.usage-cache.json`

## Environment

- claude-hud v0.0.7
- Linux (no macOS Keychain), credentials from `~/.claude/.credentials.json`

YASoftwareDev:
Bumps 'CACHE_FAILURE_TTL_MS' from 15s to 5 minutes.

With the 15s failure TTL, a single 429 response leads to an infinite cycle where the HUD retries 4 times/minute, actively preventing recovery from a transient rate-limit and keeping the 'Usage ⚠ (429)' warning permanently in the status bar.

Fixes #188, related to #187

### DIFF
--- a/src/usage-api.ts
+++ b/src/usage-api.ts
@@ -44,7 +44,7 @@ interface UsageApiResult {
 
 // File-based cache (HUD runs as new process each render, so in-memory cache won't persist)
 const CACHE_TTL_MS = 60_000; // 60 seconds
-const CACHE_FAILURE_TTL_MS = 15_000; // 15 seconds for failed requests
+const CACHE_FAILURE_TTL_MS = 300_000; // 5 minutes for failed requests (avoids 429 retry storm)
 const CACHE_LOCK_STALE_MS = 30_000;
 const CACHE_LOCK_WAIT_MS = 2_000;
 const CACHE_LOCK_POLL_MS = 50;


jarrodwatts:
Closing as superseded by merged PR #193, which carries the 429 mitigation forward on top of current main with broader cache/backoff test coverage.

PR #191, opened 2026-03-10
fix: respect retry-after header on 429 and show last successful usage during backoff

### LINKED ISSUE #188: Usage API: increase failure cache TTL to avoid 429 retry storm

Dev-RubinJo:
## Summary
- Parse `retry-after` header from 429 responses and use as cache TTL instead of fixed failure TTL
- Preserve last successful usage data in cache (`lastSuccess`) to display during 429 backoff, preventing repeated "Usage ⚠ (429)" display
- Increase fallback failure TTL from 15s → 120s for cases where `retry-after` header is absent

## Changes (`src/usage-api.ts`)
- `UsageApiResult`: added `retryAfterMs` field
- `CacheFile`: added `retryAfterMs` and `lastSuccess` fields
- `fetchUsageApi`: parse `retry-after` header on 429 responses
- `readCacheState`: use `retryAfterMs` as TTL when available; return `lastSuccess` data during backoff
- `writeCache`: accept `retryAfterMs` param; preserve `lastSuccess` from previous cache on failure
- `getUsage`: pass `retryAfterMs` to `writeCache` on API failure
- `CACHE_FAILURE_TTL_MS`: 15_000 → 120_000

## Test plan
- [ ] Verify 429 response with `retry-after: 60` header results in 60s cache TTL
- [ ] Verify last successful usage data is shown during 429 backoff period
- [ ] Verify fallback to 120s TTL when no `retry-after` header present
- [ ] Verify normal (non-429) error responses still use `CACHE_FAILURE_TTL_MS`
- [ ] Verify successful responses still use `CACHE_TTL_MS` (60s)

Fixes #188, related to #187

### DIFF
--- a/src/usage-api.ts
+++ b/src/usage-api.ts
@@ -40,11 +40,12 @@ interface UsageApiResponse {
 interface UsageApiResult {
   data: UsageApiResponse | null;
   error?: string;
+  retryAfterMs?: number;  // parsed retry-after header value in ms
 }
 
 // File-based cache (HUD runs as new process each render, so in-memory cache won't persist)
 const CACHE_TTL_MS = 60_000; // 60 seconds
-const CACHE_FAILURE_TTL_MS = 15_000; // 15 seconds for failed requests
+const CACHE_FAILURE_TTL_MS = 120_000; // 120 seconds for failed requests
 const CACHE_LOCK_STALE_MS = 30_000;
 const CACHE_LOCK_WAIT_MS = 2_000;
 const CACHE_LOCK_POLL_MS = 50;
@@ -75,6 +76,8 @@ function isUsingCustomApiEndpoint(env: NodeJS.ProcessEnv = process.env): boolean
 interface CacheFile {
   data: UsageData;
   timestamp: number;
+  retryAfterMs?: number;    // retry-after header value in ms (for 429 responses)
+  lastSuccess?: UsageData;  // last successful response to show during 429 backoff
 }
 
 interface CacheState {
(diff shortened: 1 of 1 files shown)


jarrodwatts:
Closing as superseded by merged PR #193, which is now merged on top of current main and covers the 429 handling path with the maintained implementation.

PR #192, opened 2026-03-10
fix: improve correctness and debuggability in parsing

Twodragon0:
## Summary

- **`src/git.ts`**: Replace `parseInt() || 0` with `parseInt() ?? 0` for ahead/behind count parsing. `parseInt` can legitimately return `0` (e.g., 0 commits behind), which is falsy and would be incorrectly handled by `||`. Using nullish coalescing (`??`) only falls back when the value is `NaN` (via `parseInt` returning `NaN` → `?? 0`).

  > Note: `parseInt()` returns `NaN` (not `null`/`undefined`) on failure, so `?? 0` won't catch `NaN`. However, this is consistent with the existing behavior since `NaN || 0` and `NaN ?? 0` both produce `0`. The semantic improvement is for the `0` case: `0 || 0 → 0` works by coincidence, while `0 ?? 0 → 0` works by intent.

- **`src/stdin.ts`**: Add `stderr` logging when stdin JSON parsing fails, instead of silently returning `null`. This helps users diagnose configuration issues (e.g., malformed piped input).

- **`src/transcript.ts`**: Add `stderr` logging when a JSONL transcript line fails to parse, instead of silently skipping. This aids in diagnosing transcript corruption or format changes.

## Test plan

- [x] All existing tests pass (`npm test` — 0 failures, 1 skip)
- [x] TypeScript compiles cleanly (`tsc --noEmit`)
- [ ] Manual: verify stderr messages appear only on actual parse errors, not during normal operation

### DIFF
--- a/src/git.ts
+++ b/src/git.ts
@@ -75,8 +75,8 @@ export async function getGitStatus(cwd?: string): Promise<GitStatus | null> {
       );
       const parts = revOut.trim().split(/\s+/);
       if (parts.length === 2) {
-        behind = parseInt(parts[0], 10) || 0;
-        ahead = parseInt(parts[1], 10) || 0;
+        behind = parseInt(parts[0], 10) ?? 0;
+        ahead = parseInt(parts[1], 10) ?? 0;
       }
     } catch {
       // No upstream or error, keep 0/0
--- a/src/stdin.ts
+++ b/src/stdin.ts
@@ -18,7 +18,9 @@ export async function readStdin(): Promise<StdinData | null> {
       return null;
     }
     return JSON.parse(raw) as StdinData;
-  } catch {
+  } catch (error) {
+    const message = error instanceof Error ? error.message : 'unknown error';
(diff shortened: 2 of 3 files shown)


Twodragon0:
Closing this PR — upon further review, the changes have issues:

1. **`parseInt() || 0` → `?? 0` is incorrect**: `parseInt()` returns `NaN` on failure (not `null`/`undefined`), so `NaN ?? 0` evaluates to `NaN` rather than `0`. The original `|| 0` correctly handles both `NaN` and `0` cases. This change would introduce a regression where `NaN` values slip through to the `ahead`/`behind` fields.

2. **stderr logging doesn't follow codebase conventions**: The codebase uses a `createDebug()` utility (`src/debug.ts`) that only outputs when `DEBUG=claude-hud` is set. The unconditional `process.stderr.write()` calls would cause visual noise in the statusline, especially in `transcript.ts` where malformed JSONL lines are common and expected.

Apologies for the noise.

PR #210, opened 2026-03-14
fix: stale-while-revalidate for usage API to prevent statusline timeout

jsklan:
## Summary

- When the usage cache expires, `getUsage()` blocks on keychain read (up to 3s) + HTTPS fetch (up to 15s). Claude Code kills the statusline process before it finishes, so **usage data never renders after cache expiry**.
- Adds stale-while-revalidate: when stale non-error cache data exists, the API fetch races against a 500ms timeout. Fast responses use fresh data; slow ones return stale data immediately while the fetch continues in the background updating the cache for the next render cycle.
- All existing tests pass — the race-based approach preserves exact behavior for fast API responses (as in tests) while fixing the timeout issue in production.

## Details

The change is contained to `src/usage-api.ts`:

1. New constant `SWR_TIMEOUT_MS = 500` — the max time to wait for the API before falling back to stale data
2. The fetch logic is wrapped in an inline async IIFE so it can be raced against the timeout
3. When stale non-error cache exists: `Promise.race([fetchPromise, timeoutPromise])`
   - If fetch wins → fresh data returned (same as before)
   - If timeout wins → stale data returned, fetch continues in background updating cache
4. When no stale data exists (first-ever call or cache wiped) → blocks on fetch as before

## Test plan

- [x] All 272 existing tests pass (0 failures)
- [x] TypeScript compiles cleanly with `tsc --noEmit`
- [x] Manually verified: HUD renders usage data immediately on session start instead of showing blank


jarrodwatts:
Thanks for the caching work here. I am closing this as the fix for #214 because it does not address the true cold-start path where there is no cached usage yet; it mainly helps stale-cache refresh behavior. I would rather keep the issue open than merge a change that reads as a full fix when it is not.

PR #225, opened 2026-03-17
fix: render showSpeed and showDuration in expanded layout

### LINKED ISSUE #221: `showSpeed` and `showDuration` not rendered in expanded layout
## Bug Description

`showSpeed` and `showDuration` are only implemented in the **compact** layout (`renderSessionLine()` in `session-line.ts`). When using the **expanded** layout, these config options have no effect — the values are never rendered.

## Steps to Reproduce

1. Set `lineLayout: "expanded"` in `config.json`
2. Enable `display.showSpeed: true` and `display.showDuration: true`
3. Observe that neither speed nor duration appears in the status line

## Expected Behavior

`showSpeed` and `showDuration` should be rendered in expanded layout, just as they are in compact layout.

## Root Cause

In the source code:

- `showSpeed` is only checked and rendered in `src/render/session-line.ts` (compact layout). There is no corresponding logic in any of the expanded layout renderers under `src/render/lines/`.
- `showDuration` has the same issue — only rendered in `session-line.ts`, with no expanded layout equivalent.

Other display options like `showTokenBreakdown` and `showSessionName` do have implementations in both layouts (e.g., `identity.ts` and `project.ts`), so this appears to be an oversight.

## Environment

- Plugin version: 0.0.10
- Platform: macOS (Darwin 25.3.0)

pratyush618:
## Summary

Fixes #221 — `showSpeed` and `showDuration` config options had no effect in expanded layout.

These were only implemented in compact layout (`renderSessionLine()` in `session-line.ts`). The expanded layout renderers under `src/render/lines/` had no corresponding logic.

**Change**: Added speed and duration rendering to `renderIdentityLine()` in `src/render/lines/identity.ts`, appended to the context line with `│` separators — matching the compact layout's pipe-separated style.

Expanded layout now renders:
```
Context █████░░░░░ 45% │ out: 42.1 tok/s │ ⏱️ 5m
```

## Test plan
- [x] All 272 existing tests pass
- [x] TypeScript compiles without errors
- [ ] Enable `showSpeed: true` and `showDuration: true` with `lineLayout: "expanded"` and verify both values appear on the context line


jarrodwatts:
Thanks for the contribution. I’m closing this as superseded by #232, which landed the same expanded-layout / fix on . I’d rather keep the duplicate fix path out of the history now that the bug is resolved.

PR #226, opened 2026-03-18
feat: add maxLineWidth config for explicit line wrapping

ywatanabe1989:
Add `maxLineWidth` config option that overrides automatic terminal width detection for line wrapping. When set, lines will wrap at separator boundaries (| and │) at the specified width instead of the detected terminal width.

This is useful when:
- Terminal width detection is unreliable
- Users want consistent wrapping regardless of terminal size
- The HUD is displayed in a fixed-width container

Config example: `"maxLineWidth": 80`
Default: null (use terminal width detection)

## Testing

- [ ] `npm test`
- [ ] `npm run test:coverage`

## Checklist

### DIFF
--- a/tests/render-width.test.js
+++ b/tests/render-width.test.js
@@ -1,11 +1,11 @@
-import { test } from 'node:test';
-import assert from 'node:assert/strict';
-import { render } from '../dist/render/index.js';
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { render } from "../dist/render/index.js";
 
 function baseContext() {
   return {
     stdin: {
-      model: { display_name: 'Opus' },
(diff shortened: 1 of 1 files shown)


jarrodwatts:
Thanks for the contribution. I am closing this as-is because the current implementation can widen wrapping beyond the actual terminal width, so it is not safe to merge in its present form. If someone revisits it, the override needs to be a cap, not a replacement for the detected viewport width, and the tests need to catch that case explicitly.

ywatanabe1989:
@jarrodwatts Thank you very much for your clarification! Maybe my case was just a edge case, where shell width is not reflected well! Thank you for your confirmation! Yusuke

image

PR #228, opened 2026-03-18
feat: display Claude Code version in HUD (fixes #218)

### LINKED ISSUE #218: Feature Request: Display Claude Code version in session
Claude Code iterates rapidly with frequent releases. During an active session, there is no straightforward way to check which version is currently running. Users often want to confirm whether their session is on the latest release — especially after an update.

horno1337:
Closes #218 

Adds a new 'showClaudeCodeVersion' option that displays running (C)laude (C)ode verison in the HUD (e.g. 'CC v2.1.77.')

Disabled by default, opt-in via config:
{ "display": { "showClaudeCodeVerison": true} }

20260315_16h27m23s_grim

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -7,7 +7,7 @@ export type LineLayoutType = 'compact' | 'expanded';
 
 export type AutocompactBufferMode = 'enabled' | 'disabled';
 export type ContextValueMode = 'percent' | 'tokens' | 'remaining';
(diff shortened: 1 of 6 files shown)


jarrodwatts:
Thanks for the feature proposal. I am closing this version because it shells out synchronously on the render path and ships without tests, which is too much latency and maintenance risk for a cosmetic HUD field. Happy to revisit later if there is a non-blocking/source-of-truth-based implementation.

PR #229, opened 2026-03-18
feat: add customLine config option for user-defined HUD labels

ywatanabe1989:
The existing --extra-cmd CLI argument lets users display custom labels, but requires modifying the statusLine command in settings.json. This adds a `customLine` config option that serves as a declarative alternative.

When set in config.json, it runs the specified shell command and displays the result as a dimmed label. The command must output JSON: {"label": "text"}. CLI --extra-cmd still takes priority when both are specified.

Example use case: displaying account email by resolving local credentials, since the Anthropic OAuth userinfo API is not yet available for individual users.

## Testing

- [ ] `npm test`
- [ ] `npm run test:coverage`

## Checklist

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,21 +1,28 @@
-import * as fs from 'node:fs';
-import * as path from 'node:path';
-import * as os from 'node:os';
-import { getHudPluginDir } from './claude-config-dir.js';
+import * as fs from "node:fs";
+import * as path from "node:path";
+import * as os from "node:os";
+import { getHudPluginDir } from "./claude-config-dir.js";
 
-export type LineLayoutType = 'compact' | 'expanded';
+export type LineLayoutType = "compact" | "expanded";
 
-export type AutocompactBufferMode = 'enabled' | 'disabled';
-export type ContextValueMode = 'percent' | 'tokens' | 'remaining';
-export type HudElement = 'project' | 'context' | 'usage' | 'environment' | 'tools' | 'agents' | 'todos';
+export type AutocompactBufferMode = "enabled" | "disabled";
+export type ContextValueMode = "percent" | "tokens" | "remaining";
+export type HudElement =
+  | "project"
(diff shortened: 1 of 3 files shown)


jarrodwatts:
Thanks for the work here. I merged the safer static custom-line approach in #223 instead. I do not want config.json to become a command-execution surface, so I am closing this dynamic variant rather than carrying both models forward.

ywatanabe1989:
@jarrodwatts Thank you for your consideration! I wanted to show email account and needed this. Thank you! Yusuke

image


```bash
# Resolve current Claude account email by matching subscriptionType
cld_email() {
    python3 -c "
import json, glob, os, sys

home = os.path.expanduser('~')
cred_dir = os.path.join(home, '.claude')
main_path = os.path.join(cred_dir, '.credentials.json')

if not os.path.exists(main_path):
    sys.exit(1)

email_map = {
    'ywata1989': 'ywata1989@gmail.com',
    'wyusuuke': 'wyusuuke@gmail.com',
}

main = json.load(open(main_path))
main_oauth = main.get('claudeAiOauth', {})
main_sub = main_oauth.get('subscriptionType', '')

for f in sorted(glob.glob(os.path.join(cred_dir, '.credentials-*.json'))):
    name = os.path.basename(f).replace('.credentials-', '').replace('.json', '')
    d = json.load(open(f))
    oauth = d.get('claudeAiOauth', {})
    sub = oauth.get('subscriptionType', '')
    if sub and sub == main_sub:
        print(email_map.get(name, name))
        sys.exit(0)

# Fallback: no match
" 2>/dev/null
}

# For claude-hud customLine: output JSON label
cld_email_json() {
    local email
    email=$(cld_email)
    if [[ -n "$email" ]]; then
        printf '{"label": "%s"}\n' "$email"
    else
        echo '{"label": ""}'
    fi
}
```

Issue #214, opened 2026-03-16
Usage line not shown on first launch — cold cache / API timeout


xiaodream551-a11y:
## Description

After a fresh `/claude-hud:setup`, the Usage line (5h / 7d utilization) does not appear until the usage API cache (`.usage-cache.json`) has been populated by a successful API call. On first launch, the cache is empty and the API request may not complete within the statusline render window, resulting in a missing Usage line with no visible error.

## Root Cause

The HUD is invoked as a short-lived process every ~300ms. On cold start:

1. No `.usage-cache.json` exists
2. `readKeychainCredentials` calls `execFileSync('/usr/bin/security', ...)` with a 3s timeout — on first access, macOS may prompt for Keychain permission, causing a timeout
3. If keychain fails, `recordKeychainFailure()` sets a **60-second backoff** (`KEYCHAIN_BACKOFF_MS`), suppressing all subsequent usage fetches for a full minute
4. Even if keychain succeeds, the usage API call (`fetchUsageApi`) has a 15s timeout — the statusline process may be killed before the response arrives
5. As a result, the Usage line silently doesn't render, with no indication to the user that anything is wrong

Once the cache is populated (e.g., by manually running the HUD command in a terminal), subsequent renders work perfectly.

## Steps to Reproduce

1. Fresh install: `/plugin install claude-hud` → `/claude-hud:setup`
2. Start using Claude Code — observe the HUD
3. Usage line is missing on the first session
4. Manually run the HUD command in terminal: `echo '{"model":{"display_name":"Opus"},...}' | bun src/index.ts`
5. Usage line now appears in HUD (cache was populated by step 4)

## Suggested Fix

**Pre-warm the cache during setup**: The `/claude-hud:setup` skill should trigger an initial usage API fetch after writing the config, so the cache is populated before the user starts their first session. This could be as simple as running the HUD command once with sample stdin data as part of setup Step 2/3.

Alternatively, the HUD could:
- Show a "Usage: loading..." placeholder on cold start instead of hiding the line entirely
- Spawn the API fetch in background on first render and show cached data on subsequent renders (already partially implemented, but the initial fetch timing is the issue)

## Environment

- **Platform**: macOS (Darwin, Apple Silicon M4)
- **Shell**: zsh
- **Runtime**: bun
- **Plugin version**: 0.0.10
- **Subscription**: Pro

Astro-Han:
[...]

PR #249, opened 2026-03-19
fix: usage line missing on cold start (#214)

learnabhi:
## Fix: Usage line missing on cold start

**Problem**
- On fresh install, the Usage line silently doesn't render until `.usage-cache.json` is populated
- Keychain timeout on first launch triggers a 60s backoff, blocking all subsequent usage fetches
- HUD process (~300ms lifetime) is killed before the 15s API fetch completes

**Changes**

`usage-api.ts`
- Skip keychain backoff when no `.usage-cache.json` exists — cold start must attempt a fetch regardless of prior keychain failures

Setup skill (`/claude-hud:setup`)
- Pre-warm `.usage-cache.json` at end of Step 2 by running the HUD command once with sample stdin data, before the statusline config is written

HUD render
- Show `Usage (loading...)` instead of silently hiding the line when `usageData.planName` is not yet available

**Result**
- Usage line renders immediately after setup instead of missing for up to 60s
- User gets visible feedback during the rare case where cache is still being populated

Fixes : #214

### DIFF
--- a/commands/setup.md
+++ b/commands/setup.md
@@ -171,6 +171,22 @@ Run the generated command. It should produce output (the HUD lines) within a few
 - If it hangs for more than a few seconds, cancel and debug.
 - This test catches issues like broken runtime binaries, missing plugins, or path problems.
 
++ **Pre-warm the usage cache**: After the test succeeds, run the command once with
++ sample stdin data to trigger an initial usage API fetch and populate `.usage-cache.json`:
++ 
++ **macOS/Linux**:
++ ```bash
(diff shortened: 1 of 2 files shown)


jarrodwatts:
Thanks for taking a swing at the cold-start issue. I am closing this because it does not actually solve the first-launch no-usage case in #214; it mainly changes the stale-cache path and can show a loading state even when usage is intentionally unavailable. I would rather keep #214 open for a more accurate fix.

PR #261, opened 2026-03-20
fix: fall back to cached planName when subscriptionType is null

### LINKED ISSUE #257: Usage line not shown when credentials.json has null subscriptionType
## Bug Description

The usage line does not appear in the HUD even when `display.showUsage` is set to `true` in config.

## Root Cause

In `usage-api.js`, `getUsage()` reads `subscriptionType` from `~/.claude/.credentials.json` and passes it to `getPlanName()`. When `subscriptionType` is `null`, `getPlanName()` returns `null`, causing `getUsage()` to bail out early (treating the user as an API-only user with no usage limits).

However, the user is actually an OAuth subscriber (Max plan in this case). The `subscriptionType` field in `.credentials.json` can become `null` after a token refresh — Claude Code does not always persist this field when rewriting credentials.

**Evidence:** The usage cache file (`.usage-cache.json`) correctly shows `planName: "Max"` from a previous successful fetch, confirming the user does have a valid subscription. But once the credential file loses `subscriptionType`, all subsequent fetches return `null` and the usage line disappears.

## Steps to Reproduce

1. Log in via OAuth with a Max (or Pro/Team) subscription
2. Use Claude Code until a token refresh occurs
3. Observe that `~/.claude/.credentials.json` now has `"subscriptionType": null`
4. The usage line no longer appears in the HUD

## Suggested Fix

Consider falling back to cached `planName` when `subscriptionType` is `null` but valid cached usage data exists. For example:

```js
const planName = getPlanName(subscriptionType);
if (!planName) {
    const cached = readCacheState(homeDir, now, ttls);
    if (cached?.data?.planName) {
        return cached.data;
    }
    return null;
}
```

Alternatively, the HUD could read `planName` from the cache as a fallback when credentials lack `subscriptionType`.

## Environment

- OS: Windows 11
- Claude Code: OAuth login (Max plan)
- claude-hud: v0.0.10

hoklims:
## Summary

- Fixes the usage line silently disappearing when `subscriptionType` becomes `null` in `.credentials.json` after an OAuth token refresh
- Falls back to the cached `planName` from `.usage-cache.json` instead of treating the user as an API user
- Preserves existing behavior: no cache + no subscriptionType = `null` (genuine API user)

## Root Cause

In `getUsage()`, `getPlanName(subscriptionType)` returns `null` when `subscriptionType` is empty/null, causing an early `return null` — the usage line disappears. However, the user is a valid OAuth subscriber whose `subscriptionType` was lost during a Claude Code token refresh (Claude Code does not always persist this field when rewriting `.credentials.json`).

The `.usage-cache.json` file correctly contains `planName: "Max"` (or "Pro"/"Team") from a previous successful API fetch, proving the user has a valid subscription.

## Changes

**`src/usage-api.ts`**:
- When `getPlanName()` returns `null`, check `readCacheState()` for a previously cached `planName`
- If found, use it as a display-only fallback (not auth-critical) and proceed with the API fetch
- If no cache exists either, return `null` as before (genuine API user)

**`tests/usage-api.test.js`**:
- Added test: cached planName fallback when subscriptionType becomes null (within cache TTL)
- Added test: cached planName fallback when subscriptionType is null and cache is stale (triggers fresh API fetch with cached planName)
- Existing test preserved: no cache + no subscriptionType = `null` (API user behavior unchanged)

## Test plan

- [x] `npx tsc` — compiles without errors
- [x] Manual test: 4 scenarios verified (initial Max → null subscriptionType with fresh cache → null with stale cache → no cache at all)
- [x] `npm test` — 218 passing tests, 16 pre-existing failures (Windows `execFileSync('npm')` issue in test infrastructure, unrelated to this change)

Closes #257

### DIFF
--- a/src/usage-api.ts
+++ b/src/usage-api.ts
@@ -403,14 +403,16 @@ export async function getUsage(overrides: Partial<UsageApiDeps> = {}): Promise<U
     // Determine plan name from subscriptionType
     let planName = getPlanName(subscriptionType);
     if (!planName) {
-      // subscriptionType may be null/empty after OAuth token refresh —
-      // Claude Code does not always persist this field when rewriting credentials.
-      // Fall back to cached planName if available (display-only, not auth-critical).
-      const cached = readCacheState(homeDir, now, deps.ttls);
-      if (cached?.data?.planName) {
-        planName = cached.data.planName;
-      } else {
-        // No cached plan either — genuine API user, no usage limits to show
+      // Only fall back to cache when subscriptionType is genuinely missing/empty
+      // (lost during OAuth token refresh). Explicit values like 'api' should not
+      // trigger the fallback — those users intentionally have no usage limits.
+      if (!subscriptionType.trim()) {
+        const cached = readCacheState(homeDir, now, deps.ttls);
+        if (cached?.data?.planName) {
+          planName = cached.data.planName;
+        }
+      }
+      if (!planName) {
         return null;
       }
     }
--- a/tests/usage-api.test.js
+++ b/tests/usage-api.test.js
@@ -581,6 +581,7 @@ describe('getUsage', () => {
 
     // Should return cached data (still fresh), with planName preserved
     assert.equal(afterRefresh?.planName, 'Max');
+    assert.equal(fetchCalls, 1);
   });
 
   test('falls back to cached planName when subscriptionType is null and cache is stale', async () => {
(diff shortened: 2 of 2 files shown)


jarrodwatts:
Closing as superseded by the focused maintainer restack now on main. The cached-plan fallback landed, so I do not want to keep the original PR open in parallel.

PR #264, opened 2026-03-20
Feat: non-blocking Claude Code version in HUD +tests (fixes #218)

### LINKED ISSUE #218: Feature Request: Display Claude Code version in session

horno1337:
## Summary

Replace synchronous `execSync('claude --version')` on the render path with an async `execfile` - based lookup extracted into `src/version.ts` . The module-level cache is preserved so the shell-out only happens once per process. Also deleted the stray 'version.ts' file.

Closes #218 

## Testing

- [x ] `npm test`
- [ x] `npm run test:coverage`

## Checklist

- [ x] Tests updated or not needed
- [ x] Docs updated if behavior changed

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,7 @@ Edit `~/.claude/plugins/claude-hud/config.json` directly for advanced settings s
 | `display.showAgents` | boolean | false | Show agents activity line |
 | `display.showTodos` | boolean | false | Show todos progress line |
 | `display.showSessionName` | boolean | false | Show session slug or custom title from `/rename` |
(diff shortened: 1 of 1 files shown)


jarrodwatts:
Closing as superseded by the focused maintainer restack now on main. The final landing shape stayed opt-in and narrower than the original branch, but the feature itself is now merged.

PR #278, opened 2026-03-22
feat: add contextSizeOverrides to customize max context size per model

hongdongjian:
## Summary

- Add `display.contextSizeOverrides` config option to override `context_window_size` per model
- Matching priority: exact match first, then longer substring patterns first
- Matching fields: `model.id` → `model.display_name` → `getModelName()` (normalized)
- When override is active: uses raw percent (no autocompact buffer) and auto-shows token suffix (e.g. `45% (45k/100k)`)
- In `remaining` mode with override: shows remaining tokens (e.g. `55% (55k)`)
- Rejects empty/whitespace-only keys to prevent accidental global overrides

## Config Example

```json
{
  "display": {
    "contextSizeOverrides": {
      "Opus": 150000,
      "gpt-5.4": 256000,
      "gpt": 128000
    }
  }
}
```

## Test plan

- [x] 6 config validation tests (empty keys, invalid values, non-object input)
- [x] 9 stdin matching tests (exact vs substring, case-insensitive, longer-pattern priority, model.id first)
- [x] 6 render tests (token suffix shown/hidden, raw percent with override, remaining mode)
- [x] All 179 existing tests pass (1 skipped: Windows-only)
- [x] Manual verification: `14% (55k/400k)` matches `55000/400000 = 13.75% ≈ 14%`


jarrodwatts:
Closing for now. This adds a broad config surface and changes context semantics away from Claude Code's native reporting, which does not fit the current campaign goal of keeping the default HUD simple and unsurprising. If this comes back later, it should be as a much narrower issue-backed proposal with a tighter product story.

zhenrong2019:
Hi @jarrodwatts, really hoping this feature can get merged!

**Our use case:** We run 4 Claude Code instances on a single VPS, each potentially using different models with different context windows (DeepSeek V4 Pro = 1M tokens, yet Claude Code reports 200K via stdin because it assumes Anthropic defaults). The context bar currently shows 5x inflated percentages on DeepSeek instances.

**Why per-model mapping matters over autoCompactWindow:** PR #589 (autoCompactWindow) sets a single global denominator — but when you have multiple instances using different providers simultaneously, you need per-model granularity. One instance might use DeepSeek (1M), another might use GPT-5 (256K), another Claude Opus (200K).

**What's great about this PR:** The substring-matching-per-model approach is exactly right for multi-instance, multi-provider setups. Happy to help test or contribute if there are outstanding concerns blocking the merge.

Thanks for all the work on this plugin!

PR #281, opened 2026-03-22
fix: use absolute GitHub path for marketplace plugin source

Cheng-Zi-Qing:
## Problem

The marketplace.json uses a relative path `./` for the plugin source, which Claude CLI cannot resolve when loading the marketplace. This causes the plugin installation to fail with 'Plugin not found in any configured marketplace'.

## Solution

Changed the plugin source from `./` to `jarrodwatts/claude-hud` (absolute GitHub path) so Claude CLI can properly locate and load the plugin.

## Testing

After this fix, users should be able to:
```
/plugin marketplace add jarrodwatts/claude-hud
/plugin install claude-hud
```

Without getting the installation error.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -11,7 +11,7 @@
   "plugins": [
     {
       "name": "claude-hud",
-      "source": "./",
+      "source": "jarrodwatts/claude-hud",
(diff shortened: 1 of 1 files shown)


jarrodwatts:
Closing on current evidence. I re-tested marketplace install on current main and did not reproduce the reported failure, so I do not want to merge a metadata change that is not justified by a confirmed repro. If someone can reproduce this reliably, please reopen with exact Claude Code version, logs, and install output.

PR #286, opened 2026-03-23
fix: prevent PowerShell statusline command on win32+bash

saronic-brandyn:
## Summary

On Windows with Git Bash (the most common Windows Claude Code config), the setup skill generates a PowerShell statusline command:

```
powershell -Command "& {$p=(Get-ChildItem $env:USERPROFILE\...}"
```

Claude Code invokes statusline commands through bash. Bash interprets `$p` and `$env` as empty shell variables **before** PowerShell sees them, so the command fails silently — node never starts and the HUD shows raw stdin JSON instead of rendered output.

## Changes

- Split the single "Windows" section into explicit **"Windows + Git Bash"** and **"Windows + PowerShell"** subsections
- Added a prominent warning explaining *why* PowerShell commands break under bash
- Bolded the routing table entry for `win32 + bash` to make it impossible to miss

## Root cause

The setup instructions had a routing table saying `win32 + bash → use macOS/Linux instructions`, but it was easy to skip because the PowerShell section was right below under a generic "Windows" header. The AI executing the skill would pick the PowerShell branch for any Windows platform regardless of shell.

## Test plan

- [ ] Run `/claude-hud:setup` on Windows with Git Bash shell — verify bash command is generated (not PowerShell)
- [ ] Run `/claude-hud:setup` on Windows with PowerShell shell — verify PowerShell command is still generated
- [ ] Verify the generated bash command works: `echo '{"model":{"display_name":"Opus"},...}' | node dist/index.js`

### DIFF
--- a/commands/setup.md
+++ b/commands/setup.md
@@ -100,8 +100,8 @@ This is a [Claude Code platform limitation](https://github.com/anthropics/claude
 |----------|-------|----------------|
 | `darwin` | any | bash (macOS instructions) |
 | `linux` | any | bash (Linux instructions) |
-| `win32` | `bash` (Git Bash, MSYS2) | bash (use macOS/Linux instructions) |
-| `win32` | `powershell`, `pwsh`, or `cmd` | PowerShell (use Windows instructions) |
+| `win32` | `bash` (Git Bash, MSYS2) | bash — use macOS/Linux instructions. **NEVER use PowerShell commands with bash shell.** |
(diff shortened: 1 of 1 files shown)


jarrodwatts:
Closing as superseded by the setup/docs hardening already landed on main. The useful win32 plus bash guidance from this PR was folded into the narrower maintainer-owned pass rather than merged as a separate branch.

PR #289, opened 2026-03-23
feat: performance, resilience, and Windows compatibility improvements

saronic-brandyn:
## Summary

Six improvements to claude-hud, verified against the full TypeScript source (not compiled dist/):

- **Parallel I/O** — `Promise.all` for git, usage, transcript, and config-count operations in `main()`. Cuts per-invocation latency.
- **DRY refactor** — Extracted 5 duplicated helper functions from compact/expanded renderers into `format-helpers.ts` (-172 net lines)
- **Transcript tail-read** — Files >512KB read first line + last 128KB instead of full scan. 15ms on 10k-line transcripts.
- **Transcript test suite** — 20 tests for the only module that had no dedicated test file
- **Atomic cache writes** — `atomicWriteFileSync` (write-to-temp + rename) prevents speed cache corruption from concurrent 300ms invocations
- **ASCII fallback mode** — New `display.asciiMode: true` config swaps Unicode symbols (`█░◐✓▸⏱️`) for ASCII (`#-~+>T:`) on Windows terminals with limited font support

## Test plan

- [x] `npx tsc` — zero TypeScript errors
- [x] 170/172 tests pass (2 pre-existing Windows root-path failures)
- [x] New transcript suite: 20/20 pass
- [x] Smoke test: `echo '{"model":{"display_name":"Opus"},...}' | node dist/index.js` renders correctly

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/docs/plans/2026-03-22-performance-and-resilience.md
+++ b/docs/plans/2026-03-22-performance-and-resilience.md
@@ -1,0 +1,984 @@
+# Claude HUD Performance & Resilience Improvements
+
+> **For Claude:** REQUIRED SUB-SKILL: Use superpowers:executing-plans to implement this plan task-by-task.
+
+**Goal:** Improve HUD performance, eliminate code duplication, add missing test coverage, and add ASCII fallback for Windows compatibility.
+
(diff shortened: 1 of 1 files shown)


brandyn-s:
Opened against upstream by mistake — this belongs on the fork. Closing.

Issue #293, opened 2026-03-23
Show compaction count in context bar


omrfc:
## Feature Request

During long sessions, Claude's context window fills up and gets compacted (compressed). This happens silently — users don't know how many times it occurred.

### Current

```
Context █░░░░░░░░░ 95%
```

### Proposed

Show compaction count next to context:

```
Context █░░░░░░░░░ 95%          → no compaction yet
Context █░░░░░░░░░ 95% (⟳1)    → compacted once
Context █░░░░░░░░░ 72% (⟳3)    → compacted 3 times (just compacted, back to 72%)
```

### Suggested Config

```json
{
  "display": {
    "showCompactionCount": true
  }
}
```

### Why

Compaction means earlier context is lost/compressed. Knowing how many times it happened helps users decide whether to start a fresh session (e.g., after 5+ compactions, context quality degrades significantly). It also helps debug unexpected behavior — "why did Claude forget X?" → "oh, 4 compactions happened."

omrfc:
Note: This feature should be toggleable via config — \ is already in the suggested config above. Default should probably be \ to keep the display clean for users who don't need it.

jarrodwatts:
Thanks for the request.

I dug through the current HUD inputs and implementation before deciding this one. We do not have a reliable compaction signal in the current Claude Code stdin payload or in the transcript parser today. The HUD can read context usage, rate limits, tools, agents, todos, and session naming data, but not an authoritative compaction counter or compaction event stream.

I do not want to ship a heuristic here. A guessed compaction count would drift after `/clear`, session restarts, or upstream behavior changes, and that would make the HUD look precise when it is not.

Closing this for now as blocked on upstream data exposure. If Claude Code adds an explicit compaction count or compaction event to stdin/transcript, we should reopen it and wire it in behind a config toggle.

PR #302, opened 2026-03-23
feat: enable compaction count display

harryfrzz:
## Summary
Add context compaction count display feature. Tracks when context usage drops by ≥10% (indicating compaction) and displays count as (⟳ N). Toggleable via display.showCompactionCount config option, defaults to false.

## Testing

- [x] `npm test`
- [x] `npm run test:coverage`

## Checklist

### DIFF
--- a/src/compaction.ts
+++ b/src/compaction.ts
@@ -1,0 +1,78 @@
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import * as os from 'node:os';
+
+interface CompactionState {
+  lastPercent: number | null;
+  count: number;
+  lastUpdate: number;
+}
+
+const STATE_FILE = 'compaction-state.json';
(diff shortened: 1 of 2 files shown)


jarrodwatts:
Thanks for taking this on. I'm closing the PR in its current form rather than trying to merge it with follow-up fixes.

Why I'm not taking it as-is:
- the counter is stored globally, so compaction counts can leak across unrelated sessions/projects
- the detection runs off displayed percent drops, which can miscount non-compaction events and misses some real ones
- the expanded render path mutates tracking state even when the feature is disabled
- the branch is `main`, which is not a safe branch shape for maintainership cleanup work

I do think the feature request itself is still reasonable, so I'm leaving #293 open for a better session-scoped implementation with targeted tests.

PR #304, opened 2026-03-23
feat: add per-query cost tracking to HUD

saronic-brandyn:
## Summary
- Track per-query cost deltas using a file-based cache (`.cost-cache.json`) between ~300ms statusline invocations
- Display last query cost inline: `Cost $1.42 (+$0.08) ($0.85/hr)` in both expanded and compact layouts
- Uses a 2-second settle threshold to detect when a query completes vs tool execution pauses

## Test plan
- [x] TypeScript builds clean (`npm run build`)
- [x] Smoke test with sample stdin shows query cost delta
- [ ] CI passes

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,6 +8,7 @@ import { loadConfig } from './config.js';
 import { parseExtraCmdArg, runExtraCmd } from './extra-cmd.js';
 import { getContextVelocity } from './context-velocity.js';
 import { detectCompaction } from './compaction-detector.js';
+import { getQueryCost } from './query-cost.js';
 import { getContextPercent, getBufferedPercent } from './stdin.js';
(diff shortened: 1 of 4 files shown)


jarrodwatts:
Following up here since the core request has now landed in a later maintainer PR.

We merged `display.showCost` in #374 as an opt-in offline estimate built from local transcript token usage for known Anthropic model families. That keeps the feature useful without depending on undocumented endpoints, background pricing fetches, or per-query heuristics that are hard to validate.

What did not ship from the broader earlier approaches:
- per-query delta tracking
- hourly rate output
- any cost display for unknown or Bedrock-priced sessions

So the underlying request is now addressed, but with a narrower trust-first scope.

PR #307, opened 2026-03-24
feat: add bash-output element to display recent Bash command outputs

pzheng460:
## Summary

Adds a new opt-in HUD element (`bash-output`) that displays recent Bash tool command outputs directly in the statusline. This helps users monitor command results without scrolling through conversation history — especially useful for long-running or background tasks.

**Motivation:** When Claude Code runs multiple Bash commands (particularly with `run_in_background`), their outputs get buried in the conversation. Users often miss results of commands like `sleep 600 && tail -15 /path/to/output`. This feature surfaces the latest Bash outputs in the HUD for quick visibility.

## Changes

- **`src/types.ts`** — Added `BashEntry` interface and `recentBash` field to `TranscriptData`
- **`src/transcript.ts`** — Track Bash `tool_use`/`tool_result` blocks, capture stdout/stderr output, handle serialization for transcript cache
- **`src/config.ts`** — Added `'bash-output'` to `HudElement` type, `showBashOutput` (default: `false`) and `bashOutputMaxLines` (default: `5`) to display config
- **`src/render/bash-output-line.ts`** — New render module with status icons (◐ running, ✓ success, ✗ error, ⊘ interrupted), timestamps, and truncated output
- **`src/render/index.ts`** — Wired `bash-output` into element rendering pipeline for both expanded and compact layouts

## Configuration

```json
{
  "display": {
    "showBashOutput": true,
    "bashOutputMaxLines": 5
  }
}
```

Opt-in (default `false`), consistent with `showTools`, `showAgents`, `showTodos`.

## Display Example

```
✓ [10:30:25] $ npm test
  ... 12 lines above ...
  ✔ 264 tests passed
  ✖ 1 test failed
  ℹ duration_ms 782
◐ [10:31:00] $ sleep 600 && tail -15 output.log [bg] running...
```

## Test plan

- [x] `npm run build` passes
- [x] `npm test` passes (264/265 — 1 pre-existing failure in `version.test.js`)
- [x] Verified with live Claude Code session
- [ ] Verify `showBashOutput: false` (default) shows no bash output
- [ ] Verify `showBashOutput: true` renders bash entries correctly
- [ ] Verify `bashOutputMaxLines` truncation works
- [ ] Verify transcript cache serialization/deserialization with bash entries


jarrodwatts:
Thanks for the work here. I’m closing this PR in its current form because it crosses a privacy/security boundary that I don’t want ClaudeHUD to normalize.\n\nSurfacing recent Bash stdout/stderr directly in the HUD makes it much easier to leak sensitive command output into screenshots, recordings, screen shares, and terminal history. That risk exists even when the feature is opt-in, and it is especially high for statusline UI because users stop treating it as a deliberate reveal. The HUD is also meant to stay concise; arbitrary shell output is unbounded, noisy, and can crowd out the higher-signal status information the project is built around.\n\nIf we ever revisit this space, it would need to be much narrower and safer by default, for example a minimal status-only indicator for background Bash jobs rather than raw output content.\n\nAppreciate the implementation effort. I just don’t want to merge a feature that makes accidental data disclosure easier.

PR #320, opened 2026-03-25
fix: prevent invalid JSON escape in statusLine command

### LINKED ISSUE #315: Files with errors are skipped entirely, not just the invalid settings.
***

## Summary
The `claude-hud` plugin writes invalid JSON syntax to `~/.claude/settings.json` when configuring the `statusLine.command`. Specifically, it uses `\$` (single escape) instead of `\\$` (double escape) for `awk` variables, which causes Claude Code to crash on startup with an "Invalid or malformed JSON" error.

## Steps to Reproduce
1. Install or enable the `claude-hud` plugin (e.g., `jarrodwatts/claude-hud`).
2. The plugin automatically writes to `~/.claude/settings.json` and updates the `statusLine` configuration.
3. Open `~/.claude/settings.json` and observe the `command` string under `statusLine`. It contains invalid JSON escape sequences like `\$(NF-1)` and `\$0`.
4. Run `claude` in the terminal.
5. See the "Invalid or malformed JSON" startup error.

## Expected Behavior
The plugin should properly double-escape the dollar signs when writing bash commands into the JSON file. It should write `\\$(NF-1)` and `\\$0` so that it passes JSON validation and is correctly parsed as `\$` by the underlying shell, allowing Claude Code to start normally.

## Actual Behavior
The plugin writes `\$(NF-1)` and `\$0` into the JSON string. Since `\$` is not a valid JSON escape sequence, it breaks the entire `settings.json` file. Claude Code fails to launch and prompts: `Settings Error: Invalid or malformed JSON`.

## Environment
- **OS:** macOS
- **Node/Bun version:** Node v24.14.0
- **Claude Code version:** *2.1.81 (Claude Code)*

## Logs or Screenshots
**Error Output in Terminal:**
```text
─────────────────────────────────────────────────────────────────
  Settings Error     

  /Users/yh/.claude/settings.json
   └ Invalid or malformed JSON
                                                                                                                                                                                                                                                                                       
  Files with errors are skipped entirely, not just the invalid settings.

  ❯ 1. Exit and fix manually                                  
    2. Continue without these settings
               
  Enter to confirm · Esc to cancel
```

**Problematic JSON snippet generated by the plugin:**
```json
"statusLine": {
  "type": "command",
  "command": "bash -c 'plugin_dir=$(ls -d \"${CLAUDE_CONFIG_DIR:-$HOME/.claude}\"/plugins/cache/claude-hud/claude-hud/*/ 2>/dev/null | awk -F/ \"{ print \$(NF-1) \\\"\\t\\\" \$0 }\" | sort -t. -k1,1n -k2,2n -k3,3n -k4,4n | tail -1 | cut -f2-); exec \"/Users/yh/.nvm/versions/node/v24.14.0/bin/node\" \"${plugin_dir}dist/index.js\"'"
}
```

***

bluzername:
## Problem

When running `/claude-hud:setup`, the setup step that writes the command into `settings.json` sometimes produces invalid JSON because `$` characters from the awk expression (like `$(NF-1)` and `$0`) get escaped as `\$`. But `\$` is not valid JSON escape sequence, so the whole `settings.json` file becomes unparseable and Claude Code shows the "Invalid or malformed JSON" error on startup.

I was getting this exact error after fresh setup and it took me some time to figure out what happened. The settings file was corrupted by the escaped dollar signs.

## What I changed

Added explicit guidance in `commands/setup.md` Step 3 about JSON escaping rules. The key points:
- Dollar sign `$` does not need escaping in JSON strings, it is valid as-is
- Manually escaping as `\$` breaks the file because only `\"`, `\\`, and control characters are valid JSON escapes
- After writing should verify the output is valid JSON

This is a prompt-level fix since `setup.md` is the instruction file that guides Claude during the setup process. The root cause is Claude sometimes "helpfully" escaping `$` when it shouldn't.

## How to test

1. Run `/claude-hud:setup` on fresh install
2. Check `~/.claude/settings.json` - the `statusLine.command` value should contain raw `$` characters, not `\$`
3. Claude Code should start normally without JSON error

Closes #315

### DIFF
--- a/commands/setup.md
+++ b/commands/setup.md
@@ -214,6 +214,12 @@ Read the settings file and merge in the statusLine config, preserving all existi
 If the file doesn't exist, create it. If it contains invalid JSON, report the error and do not overwrite.
 If a write fails with `File has been unexpectedly modified`, re-read the file and retry the merge once.
 
+**CRITICAL - JSON escaping**: The generated command contains `$` characters (from bash variables and awk expressions like `$(NF-1)` and `$0`). When writing to JSON:
+- Do NOT manually escape `$` as `\$` - this is NOT a valid JSON escape sequence and will break `settings.json` parsing entirely (#315).
+- Instead, use the Read tool to load the existing settings, construct the merged object in your working memory, and use the Edit or Write tool to produce valid JSON. The `$` character is valid as-is inside JSON strings and does not need any escaping.
+- Only these characters require escaping in JSON strings: `"` (as `\"`), `\` (as `\\`), and control characters (`\n`, `\t`, etc.).
+- After writing, verify the output is valid JSON by reading it back.
+
 ```json
 {
   "statusLine": {


jarrodwatts:
Thanks for the fix and the clear write-up. The setup JSON-escaping guidance has already landed on  via #360 while I was working through the issue queue, so I’m closing this as superseded to avoid leaving duplicate PRs open. The contribution was directionally correct and appreciated.

PR #324, opened 2026-03-26
Fix getBufferedPercent to prefer native percentage from Claude Code v2.1.6+

nmurrell07:
The getBufferedPercent function now prefers the native bufferedPercent field from Claude Code v2.1.6+, which matches the HUD's own context output. The fallback manual calculation with buffer is only used for older versions that don't provide the native percentage. This ensures the HUD percentage matches what Claude Code displays in its context window, fixing the discrepancy where the HUD showed a different percentage than the actual context usage.


gvago:
Hey @jarrodwatts , I'd like to propose using Qodo's ffoss - https://www.qodo.ai/solutions/open-source/ 
https://github.com/marketplace/qodo-merge-pro-for-open-source
WDYT?

Issue #295, opened 2026-03-23
Show agent/skill names instead of generic 'Agent' and 'Skill'


omrfc:
## Feature Request

Currently, active and completed agents/skills show generic labels:

```
◐ Agent | ✓ Agent ×5 | ✓ Skill ×1
```

### Proposed

Show the actual agent type and skill name:

```
◐ gsd-executor [opus]: Executing P85 wave-1 (5m 12s)
✓ Explore [haiku] (12s) | ✓ gsd-executor [opus] (5m) | ✓ claude-code-guide [sonnet] (32s)
✓ Skill: gsd:execute-phase (5m) | ✓ Skill: commit (8s)
```

### Suggested Config

```json
{
  "display": {
    "agentDetail": "name",
    "skillDetail": "name"
  }
}
```

- `"minimal"` — current behavior (generic "Agent"/"Skill")
- `"name"` — show actual agent type and skill name

### Why

When multiple agents run, `◐ Agent` tells nothing about what's happening. Showing the name (e.g., `gsd-executor`, `Explore`, `code-reviewer`) gives instant clarity.

This is a prerequisite for more advanced features like subagent hierarchy (#291) and per-agent context display.

_Splits out the core naming feature from #291._

jarrodwatts:
Closing this as largely resolved on current main.

For agents, ClaudeHUD already renders the actual agent type plus model, description, and elapsed time instead of a generic `Agent` label.

I am not carrying the "Skill" part forward here because there still is not a stable first-class skill data model in the HUD pipeline to render in the same trustworthy way. If a concrete skill metadata source shows up later, that would be better handled as a narrower follow-up than keeping this broader issue open.

PR #328, opened 2026-03-26
feat: show agent/skill names instead of generic labels (#295)

sunyj-003:
## Testing

- [ ] `npm test`
- [ ] `npm run test:coverage`

## Checklist

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -7,7 +7,9 @@ export type LineLayoutType = 'compact' | 'expanded';
 
 export type AutocompactBufferMode = 'enabled' | 'disabled';
 export type ContextValueMode = 'percent' | 'tokens' | 'remaining' | 'both';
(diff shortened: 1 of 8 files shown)


jarrodwatts:
Thanks for the contribution. I’m closing this PR in its current form because it ends up being broader than issue #295 and I don’t think the added  surface is proven enough to merge yet.\n\nThe agent-label part is directionally good. The blocker is the rest of the scope around it:\n- new transcript parsing for a  tool shape\n- a new  HUD element and render path\n- new config surface (, ,  in elementOrder)\n- transcript cache/schema expansion\n\nI checked local Claude transcript data on this machine before deciding, and I could not find real  tool entries to validate that parser assumption. Right now that means we’d be merging synthetic support for an internal transcript shape we haven’t actually confirmed in the wild.\n\nIf this comes back, the mergeable path is likely narrower:\n- either a focused agent-name improvement only, or\n- a skills implementation backed by verified transcript examples plus tighter behavior tests\n\nI’m leaving #295 open rather than closing the request itself.

PR #337, opened 2026-03-30
fix: use vm_stat for accurate macOS memory percentage

arksou:
os.freemem() on macOS only returns Mach "Pages free", ignoring inactive/purgeable pages that macOS uses as file cache. This caused the HUD to show ~100% RAM usage. Now uses vm_stat to compute Used = (Active + Wired) * pageSize, with fallback to os.freemem().

## Testing

- [ ] `npm test`
- [ ] `npm run test:coverage`

## Checklist


jarrodwatts:
Thanks for the fix here. I carried it forward in #367 and merged that follow-up instead of merging this PR directly.

Why the follow-up branch was needed:
- this PR branch was `main`, which isn't a safe branch shape for maintainer cleanup work
- the original PR included generated `dist/` output, and I kept the merged branch source-only

The merged fix is here: #367

PR #340, opened 2026-03-30
feat: add showAccount display option

charanganesh:
## Summary

- Adds a `display.showAccount` config option that shows the currently logged-in account (email, display name, and/or organization) in the project/session line
- Disabled by default — users opt in via `config.json`
- Works in both compact and expanded layouts
- Color customizable via `colors.account` (defaults to dim)
- Reads `oauthAccount` from `~/.claude.json` with mtime-based caching to avoid re-parsing on every 300ms tick

## How it looks

With `{ "display": { "showAccount": true } }` in config:

- **Distinct org:** `[Opus] │ Alice @ Acme Corp`
- **Personal org:** `[Opus] │ user@example.com`

## Test plan

- [x] `npm test` — 17 new tests, all 281 existing tests pass
- [x] `npm run test:coverage`
- [x] Tested manually in both compact and expanded layouts
- [x] Verified mtime caching works (file only re-read when changed)
- [x] No `dist/` changes included

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to Claude HUD will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- New `display.showAccount` option to display the currently logged-in account (email or display name with organization) in the project/session line. Disabled by default; customizable via `colors.account`.
+
 ### Changed
 - Simplify usage display to rely only on Claude Code's official stdin `rate_limits` fields.
 - Remove the background OAuth usage API fallback, related cache/lock behavior, and credential-derived subscriber plan labels from the HUD.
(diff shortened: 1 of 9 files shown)


jarrodwatts:
Thanks for the careful implementation here. I’m closing this PR in its current form because the data source is outside the project’s supported trust boundary.\n\nClaudeHUD is trying to stay aligned with Claude Code’s documented statusline contract, which is the JSON passed on stdin. Reading account identity from ~/.claude.json is not part of that contract, can drift from the active runtime session, and also raises a privacy risk for screenshots/screen shares even when opt-in. We recently tightened this area so the HUD only shows auth/provider labels when they can be positively identified from runtime input, not inferred from side-channel auth/session files.\n\nIf we add this class of feature later, the safer path would be either:\n- a manual user-configured label, or\n- an official stdin field from Claude Code that exposes account identity\n\nAppreciate the work here. The implementation is tidy; I just don’t want to merge it on top of an unsupported and privacy-sensitive source of truth.

PR #342, opened 2026-03-31
feat: add claude server status indicator

kimeloo:
## Summary

Add a real-time Claude Code service status indicator to the HUD's first line.

- Fetches Claude Code component status from the `status.claude.com` API (`/api/v2/summary.json`)
- Displays a green dot (`●`) when operational; expands to show status icon + active incident title on degraded/outage/maintenance states
- Uses inline fetch with file-based caching (5-minute TTL), following the same pattern as `version.ts`
- Opt-in via `"showStatus": true` in config

**Display examples:**
[Opus] │ my-project git:(main*) │ ●                              ← operational
[Opus] │ my-project git:(main*) │ ◐ Elevated errors on Opus 4.6  ← degraded (yellow)
[Opus] │ my-project git:(main*) │ ✖ Claude Code service down     ← major outage (red)

**example1: operational**
image

**example2: degraded**
image

**example3: outage**
image

## Testing

- [ ] `npm test`
- [ ] `npm run test:coverage`
- Manually verified all status states (operational, degraded, partial/major outage, maintenance) by injecting test cache data

## Checklist

- [x] Tests updated or not needed — no tests added; status fetch relies on external API, manual verification done with injected cache data

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ Claude HUD gives you better insights into what's happening in your Claude Code s
 | **Tool activity** | Watch Claude read, edit, and search files as it happens |
 | **Agent tracking** | See which subagents are running and what they're doing |
 | **Todo progress** | Track task completion in real-time |
+| **Service status** | See Claude Code service health at a glance — know if an outage is you or them |
 
 ## What You See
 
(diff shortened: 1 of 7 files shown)


jarrodwatts:
Thanks for the PR. I’m closing this version because it adds an external network dependency to the HUD render path, and that is too risky for a statusline process even behind an opt-in flag.\n\nThe main blockers are:\n- stale or cold-cache renders can block on a remote request\n- multiple Claude sessions can dogpile the status endpoint at cache expiry\n- the current incident mapping can show the wrong incident text\n- there are no targeted tests for timeout/error/schema-drift/cache behavior\n\nIf this comes back later, the acceptable shape would be much narrower: cache-first, non-blocking, short-timeout best-effort refresh only, and solid tests around cache hits/misses/errors. I don’t want to merge a feature that can slow or destabilize the HUD itself.\n\nAppreciate the direction and the write-up. This is a reasonable idea, but it needs a stricter implementation boundary than this PR uses.

kimeloo:
Understood. Thanks for taking the time to review and for pointing out the potential stability issues with this approach. I'll keep those constraints in mind for future contributions!

PR #349, opened 2026-04-01
feat: add Chinese (zh) language support

### LINKED ISSUE #334: Add Chinese (中文) language support

xiaodream551-a11y:
## Summary
- add lightweight i18n infrastructure under `src/i18n/` with no runtime dependencies
- add explicit `language` config support with English as the default and `zh` as an opt-in
- translate user-facing HUD labels and status text while keeping technical terms like `CLAUDE.md`, `MCPs`, `git`, and `CC v` unchanged
- add render coverage for Chinese labels, including a narrow-terminal width check
- merge current `main` into the branch so the PR keeps recent fixes like model formatting and the auth-label trust fix

## Configuration

Users can opt into Chinese labels in `~/.claude/plugins/claude-hud/config.json`:

```json
{
  "language": "zh"
}
```

If no language is set, Claude HUD stays in English.

## Testing
- npm run build
- node --test tests/i18n.test.js tests/render-width.test.js tests/render.test.js tests/index.test.js tests/integration.test.js
- npm test (matches current baseline except for the pre-existing `tests/version.test.js` cache failure on main)

Closes #334

### DIFF
--- a/tests/render.test.js
+++ b/tests/render.test.js
@@ -1018,7 +1018,7 @@ test('renderSessionLine shows 7d reset countdown in text-only mode', () => {
 
   const line = stripAnsi(renderSessionLine(ctx));
   assert.ok(line.includes('7d: 85%'), `should include 7d label and percentage: ${line}`);
-  assert.ok(line.includes('(1d 4h)'), `should include 7d reset countdown in text-only mode: ${line}`);
+  assert.ok(line.includes('(resets in 1d 4h)'), `should include 7d reset countdown in text-only mode: ${line}`);
 });
 
 test('renderSessionLine respects sevenDayThreshold override', () => {


jarrodwatts:
thanks @xiaodream551-a11y - i think this is a great idea since the chinese community of users has grown massively! will look into adding support for this

jarrodwatts:
Thanks again for shipping this. This is the right long-term direction for the project.

I want to make the maintainer stance explicit now that the board work is caught up:
- English stays the default for the simplest onboarding path
- Chinese support is first-class and intentionally supported via explicit `language: "zh"`
- technical tokens like `CLAUDE.md`, `MCPs`, git refs, and version labels should stay stable rather than being over-localized

For follow-up scope, I want to build on this foundation carefully: prioritize Chinese install/config/troubleshooting docs and keep validating CJK width/render behavior so the HUD stays clean in real terminal usage.

This PR was a strong starting point and the current i18n path should grow from here rather than being replaced.

PR #358, opened 2026-04-03
feat: context intelligence — cost, cache ratio, efficiency colors

gpalmgg:
## Summary

- **Session cost display**: Shows `$1.40` from Claude Code's `cost.total_cost_usd` (already piped via stdin but unused)
- **Cache efficiency ratio**: Shows `85% cached` computed from `cache_read_input_tokens / total`
- **Token count**: Always shows `(160k)` next to the context percentage for grounding
- **Efficiency-aware colors**: Cost and cache values shift green → yellow → red based on a composite zone (context% + cache ratio), not just raw context percentage
- **Adaptive reclaimable estimate**: Shows `🧹 ~40k reclaimable` only when stale tool results exceed a configurable threshold

### Why

The context bar currently shows `Context ████░░░░░░ 16%` — which tells you how full the window is but nothing about cost or efficiency. At 16% of a 1M window, you're already carrying 160k tokens. Whether that's cheap (90% cached) or expensive (cache broken after compact) is invisible.

This surfaces the data Claude Code already sends (`cost.total_cost_usd`, `cache_read_input_tokens`) that the plugin wasn't using.

### New config options

All default to `false` for backward compatibility:

| Config key | Type | Default | Description |
|---|---|---|---|
| `display.showCost` | boolean | false | Show session cost in USD |
| `display.showCacheRatio` | boolean | false | Show cache hit percentage |
| `display.showReclaimable` | boolean | false | Show reclaimable token estimate |
| `display.reclaimableThreshold` | number | 30000 | Min tokens before showing reclaimable |

### Output examples

```
# Green zone — efficient
Context ████░░░░░░ 16% (160k) │ $1.40 │ 85% cached

# Yellow zone — getting heavy
Context █████░░░░░ 50% (500k) │ $4.20 │ 50% cached

# Red zone — compact would help
Context ███████░░░ 70% (600k) │ $8.50 │ 33% cached
```

### Efficiency zone logic

| Condition | Color |
|---|---|
| Context <40% OR (Context <60% AND cache >80%) | Green |
| Context 40-60% AND cache <80% OR Context 60-80% | Yellow |
| Context >60% AND cache <70% OR Context >80% | Red |

## Test plan

- [x] All 264 existing tests pass (1 pre-existing flaky in version.test.js)
- [x] Updated integration test fixture for token count in context line
- [x] Tested green/yellow/red zones with synthetic stdin data
- [x] Verified backward compatibility — new features default to off
- [x] Verified config merge preserves existing user settings

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -35,8 +35,10 @@ Claude Code → stdin JSON → parse → render lines → stdout → Claude Code
 
 **Native from stdin JSON** (accurate, no estimation):
 - `model.display_name` - Current model
-- `context_window.current_usage` - Token counts
+- `context_window.current_usage` - Token counts (input, output, cache_creation, cache_read)
+- `context_window.total_input_tokens` / `total_output_tokens` - Cumulative session tokens
 - `context_window.context_window_size` - Max context
+- `cost.total_cost_usd` - Session cost in USD (calculated by Claude Code using internal pricing)
 - `transcript_path` - Path to session transcript
 
 **From transcript JSONL parsing**:
@@ -87,7 +89,7 @@ src/
 
 ```
 [Opus] │ my-project git:(main*)
-Context █████░░░░░ 45% │ Usage ██░░░░░░░░ 25% (1h 30m / 5h)
+Context █████░░░░░ 45% (450k) │ $3.20 │ 78% cached │ Usage ██░░░░░░░░ 25% (1h 30m / 5h)
 ```
 
 Lines 1-2 always shown. Additional lines are opt-in via config:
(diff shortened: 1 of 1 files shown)


Yimaha:
might be a good idea to remove all the changes under dist/*. its generated after compile, and is suppose to only be in the release process not the pr (its making pr very messy)

`  git update-index --assume-unchanged dist/*` something alone this line would help, once that's ran the repo assumes there is no change in dist/ folder

jarrodwatts:
Thanks for the work here. I’m closing this PR in its current form because it bundles together a few different ideas, and some of them fall below the trust bar for HUD metrics.\n\nThe main issue is that this mixes authoritative runtime data with heuristic interpretation:\n-  is grounded if Claude Code is already sending it\n- a raw cache ratio can be defensible if it is presented plainly\n- the reclaimable estimate is not grounded enough to ship as HUD truth\n- the composite efficiency coloring changes the meaning of existing context colors in a way that is hard to reason about\n\nFor this project, misleading metrics are a bug. If the HUD says , users will treat that as a real number, not as a guess derived from stale transcript state. I don’t want to merge that kind of heuristic without an authoritative upstream signal.\n\nThis PR also includes generated  changes, and it would be easier to review if the ideas were split up. The mergeable path here is narrower:\n- a focused cost display using official stdin/runtime fields\n- possibly a plain cache ratio display, if it stays raw and well-tested\n- no reclaimable estimate unless Claude exposes a real signal for it\n\nI’m leaving related issues open where the underlying request still makes sense, but I don’t want to merge this combined version as-is.

Yimaha:
damn I really like the price idea, let me just try to get something out for the price tracker since I don't think its a feature yet

jarrodwatts:
Thanks again for pushing on this.

The cost request did land now, but through a narrower path in #374. I kept the merge scoped to an opt-in offline estimate derived from local transcript token usage for known Anthropic model families, with the estimate hidden for Bedrock and unknown pricing cases.

I did not carry forward the reclaimable estimate or the composite efficiency coloring. For this HUD, anything that looks authoritative but is actually heuristic is a trust bug, so I want those features to wait for a stronger upstream signal.

The end result is:
- `display.showCost` shipped
- no undocumented API/OAuth usage
- no network pricing fetches
- no `dist/` in the PR path

If Claude Code exposes a stronger runtime signal later, I’m open to revisiting the raw cache-side metrics in a smaller follow-up.

PR #374, opened 2026-04-04
feat: add offline estimated cost display

### LINKED ISSUE #305: Feature request: Display estimated cost based on token usage and model pricing
### Description

Add a `display.showCost` option that calculates and displays the estimated cost of the current session based on actual token usage (input, cache, output) and the model's pricing.

### Motivation

When using expensive models like Opus 4.6, it's very useful to see real-time cost tracking directly in the HUD. The HUD already tracks token counts (`input_tokens`, `cache_creation_input_tokens`, `cache_read_input_tokens`, `output_tokens`) — the missing piece is multiplying these by the model's per-token pricing.

### Suggested Implementation

- New config option: `display.showCost: true/false`
- Display format (example): `$0.42 (in: $0.30 + cache: $0.02 + out: $0.10)`
- Or a simpler format: `Cost: $0.42`
- Pricing could be looked up from a built-in table keyed by model name, or fetched from the Anthropic API pricing endpoint
- Should update in real-time as tokens accumulate during the session

### Example Pricing Table

| Model | Input (per 1M) | Cache Write (per 1M) | Cache Read (per 1M) | Output (per 1M) |
|-------|----------------|----------------------|----------------------|------------------|
| Opus 4.6 | $15 | $18.75 | $1.50 | $75 |
| Sonnet 4.6 | $3 | $3.75 | $0.30 | $15 |
| Haiku 4.5 | $0.80 | $1.00 | $0.08 | $4 |

### Config Example

```json
{
  "display": {
    "showCost": true
  }
}
```

This would be a valuable addition for users who want to monitor their API spend at a glance.

jarrodwatts:
## Summary
- add an opt-in `display.showCost` toggle for offline session cost estimates
- estimate from local transcript token usage only for known Anthropic model families
- hide the estimate for Bedrock and unknown model/pricing cases, and add English/Chinese labels

## Why
Closes #305 without adding network lookups or undocumented API dependencies. The value is intentionally labeled as an estimate and stays hidden when pricing cannot be matched confidently.

## Verification
- `npm run test:coverage`

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,7 @@ Chinese HUD labels are available as an explicit opt-in. English stays the defaul
 | `display.showContextBar` | boolean | true | Show visual context bar `████░░░░░░` |
 | `display.contextValue` | `percent` \| `tokens` \| `remaining` \| `both` | `percent` | Context display format (`45%`, `45k/200k`, `55%` remaining, or `45% (45k/200k)`) |
 | `display.showConfigCounts` | boolean | false | Show CLAUDE.md, rules, MCPs, hooks counts |
+| `display.showCost` | boolean | false | Show an offline estimated session cost from transcript token usage for known Anthropic model families |
 | `display.showOutputStyle` | boolean | false | Show the active Claude Code `outputStyle` from settings files as `style: <name>` |
 | `display.showDuration` | boolean | false | Show session duration `⏱️ 5m` |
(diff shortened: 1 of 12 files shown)


Yimaha:
@jarrodwatts  why not use cost.total_cost_usd? is it not accurate?

PR #381, opened 2026-04-04
feat: use cost.total_cost_usd from stdin instead of manual token estimation

### LINKED ISSUE #380: feat: use cost.total_cost_usd from stdin instead of manual token estimation
## Summary

The current cost display (added in #374, closes #305) manually estimates session cost by multiplying token counts from the transcript against a hardcoded pricing table. This approach has known limitations:

- The pricing table goes stale when Anthropic adjusts pricing
- Returns `null` for non-Anthropic model IDs it doesn't recognise
- Labels the value as an **estimate** because it is one

Claude Code already provides a native `cost.total_cost_usd` field in the stdin JSON, updated after every API response. This is the same value Claude Code tracks internally — no estimation needed.

## Proposed Change

Replace `estimateSessionCost()` in `src/cost.ts` with a direct read of `stdin.cost?.total_cost_usd` in `src/index.ts`.

- Keep the existing display layer (`renderCostEstimate` in `src/render/lines/cost.ts`, session-line, project-line) structurally unchanged
- Remove the hardcoded `ANTHROPIC_MODEL_PRICING` table and `estimateSessionCost` logic
- Change the label from `"Est."` → `"Cost ($USD)"` to reflect that the value is accurate and denominated in USD
- Handle `null`/absent gracefully — the field is absent before the first API response in a session

## Known unknowns

The official statusline docs describe `cost.total_cost_usd` as "Total session cost in USD" without specifying provider-specific behavior. For Bedrock or Vertex-routed sessions, Claude Code may report `$0.00` or omit the field entirely since billing is handled by the cloud provider rather than Anthropic directly. This needs verification — the display will simply stay hidden if the field is absent or zero.

## Benefits

- Accurate for standard API sessions — matches what Claude Code itself tracks
- Simpler code: no pricing table to maintain, no model name pattern matching

## Reference

- Original issue: #305
- PR that added estimation: #374
- Claude Code statusline docs: https://code.claude.com/docs/en/statusline (`cost.total_cost_usd` — "Total session cost in USD")

Yimaha:
## Summary

Closes #380

Replaces the hardcoded Anthropic pricing table and `estimateSessionCost()` logic with a direct read of `stdin.cost.total_cost_usd` — the native field Claude Code provides after each API response. The display layer is kept structurally the same; only the data source changes.

## Changes

- **`src/types.ts`** — add `cost.*` fields to `StdinData`; replace `CostData` interface with a plain `costUsd: number | null` on `RenderContext`
- **`src/index.ts`** — extract `stdin.cost.total_cost_usd`, validate finite number, pass as `costUsd`
- **`src/cost.ts`** — delete dead estimation logic (pricing table, `estimateSessionCost`, helpers); keep only `formatUsd`
- **`src/render/lines/cost.ts`** — read `ctx.costUsd` directly instead of calling `estimateSessionCost`
- **`src/i18n/`** — rename key `label.estimatedCost` → `label.cost`; update label to `"Cost ($USD)"` / `"耗油量（美元）"` (zh)
- **`tests/core.test.js`** — remove three `estimateSessionCost` tests that no longer apply
- **`tests/index.test.js`** — add two tests: `costUsd` populated from stdin, `costUsd` null when field absent
- **`tests/render.test.js`** — rewrite three cost render tests to set `ctx.costUsd` directly; add `costUsd: null` to `baseContext()`
- **`README.md`** — update `showCost` description and explanation

## Known unknowns

The docs describe `cost.total_cost_usd` as "Total session cost in USD" without specifying behavior for Bedrock or Vertex-routed sessions. Since billing in those cases goes through AWS/GCP rather than Anthropic, Claude Code may report `$0.00` or omit the field. The display stays hidden when `costUsd` is `null`, so worst case is the field doesn't appear for those users — same behavior as before. Needs real-world verification.

## Test results

```
ℹ tests 348
ℹ pass 347
ℹ fail 0
```

## Ref display
```
  [Sonnet 4.6]
  Work/claude-hud git:(feat/native-cost...
  synchronous-weaving-sifakis │ CC v2.1.92
  ⏱️   40m │ Cost ($USD) $3.13
  Context ██████░░░░ 55% (110k/200k)
  Usage ███░░░░░░░ 26% (resets in 3h 26m)
  Weekly ██░░░░░░░░ 20% (resets in 6d 4h)
  1 CLAUDE.md
  ✓ Bash ×14 | ✓ Read ×3 | ✓ Edit ×2
  ✓ Write ×1
  ✓ claude-code-guide: Search Claude Co...
  ⏵⏵ accept edits on (shift+tab to cycle)
```


jarrodwatts:
I pushed a maintainer follow-up commit to this branch to tighten the implementation before merge.

What changed:
- prefer Claude Code's native `cost.total_cost_usd` field when it is present
- keep the existing transcript-based estimate as a fallback for direct Anthropic sessions when the native field is absent or invalid
- keep cost hidden for routed providers such as Bedrock instead of showing a misleading `$0.00`
- fix the Chinese cost label and align the README with the shipped behavior
- add regression coverage for native cost, fallback estimation, and provider-routed hiding

Verification:
- `npm run build && node --test tests/render.test.js tests/index.test.js tests/core.test.js`
- `npm test`

This keeps the original PR and author history intact while raising the trust/safety bar for the cost display.

PR #387, opened 2026-04-05
fix: honor COLUMNS override before stream-based width detection

### LINKED ISSUE #385: [Bug] Windows: terminal width detected as ~50% of actual width, add            terminalWidth config override
Environment                             
  - OS: Windows 11                             
  - Claude Code HUD: v0.0.12                                                         
  - Terminal: Windows Terminal                                                       
                                               
  Problem                                                                            

  Root Cause

  process.env.COLUMNS:

  const cols = stdoutCols > 0

   fallback is never reached.

  Proposed Fix

  process.stdout.columns)

  Feature Request

Image

saschabuehrle:
## Summary
- prioritize COLUMNS as an explicit terminal width override in width detection
- apply the same precedence to adaptive progress-bar width logic
- update width-related tests to cover the override behavior

## Why
In some Windows subprocess contexts, stdout columns can be incorrect. Respecting COLUMNS first lets users enforce the correct width and avoids truncated HUD output.

## Testing
- npm run build
- node --test tests/terminal.test.js tests/render-width.test.js

Closes #385

Greetings, saschabuehrle

### DIFF
--- a/src/render/index.ts
+++ b/src/render/index.ts
@@ -30,21 +30,23 @@ function stripAnsi(str: string): string {
 }
 
 function getTerminalWidth(): number {
+  // Respect an explicit COLUMNS override first. This is useful when stdout/stderr
+  // columns are inaccurate (for example some Windows subprocess scenarios).
+  const envColumns = Number.parseInt(process.env.COLUMNS ?? '', 10);
+  if (Number.isFinite(envColumns) && envColumns > 0) {
+    return envColumns;
+  }
+
   const stdoutColumns = process.stdout?.columns;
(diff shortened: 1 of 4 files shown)


mynameistito:
wait so with this PR it will change the behaviour of this being stacked on one another? 
image

saschabuehrle:
No stacking behavior change in this patch. It only changes width detection priority so COLUMNS is respected when it is explicitly set. Layout/stacking logic is untouched.\n\nGreetings, saschabuehrle

jarrodwatts:

PR #388, opened 2026-04-05
fix(render): skip line wrapping when terminal width is unreliable

TravelThePath:
## Summary

- **`getTerminalWidth()` now returns `null`** instead of falling back to `UNKNOWN_TERMINAL_WIDTH` (40) when no reliable width source is available (`stdout.columns`, `stderr.columns`, `COLUMNS` env var all undefined)
- **Wrapping is skipped when width is `null`** — the parent process (Claude Code's Ink layer with `wrap=truncate`) handles truncation instead of the HUD doing it with a wrong value
- **`renderExpanded()` no longer combines Context/Usage lines** when width is unknown, keeping them on separate lines as the safe default

## Problem

When running as a statusline subprocess, stdout and stderr are piped — `process.stdout.columns`, `process.stderr.columns`, and `process.env.COLUMNS` are all `undefined`. The previous code fell back to `UNKNOWN_TERMINAL_WIDTH = 40`, causing `wrapLineToWidth()` to split compact session lines at `|` separators even when the actual terminal was 120+ columns wide.

This is a known Claude Code limitation: [anthropics/claude-code#22115](https://github.com/anthropics/claude-code/issues/22115) — the statusline subprocess has no way to detect terminal width since Claude Code doesn't pass `COLUMNS` to child processes.

### Before (compact mode, 120-col terminal)

```
[Opus 4.6 (1M context)] ████░░░░░░ 4%
Github | keep it simple
▶▶ auto mode on (shift+tab to cycle)
```

The session line wraps at 40 chars, pushing `Github | keep it simple` to a second line.

### After

```
[Opus 4.6 (1M context)] ████░░░░░░ 4% | Github | keep it simple
▶▶ auto mode on (shift+tab to cycle)
```

Single line as intended. Claude Code's Ink layer truncates if it actually exceeds terminal width.

## Approach

Rather than trying to detect width via `tput cols` or other workarounds (which fail on Windows, add subprocess overhead, or return stale values), the fix follows the principle: **if you don't know the width, don't wrap**. The caller (Claude Code) owns the terminal and is the right place to handle truncation.

When a reliable width source IS available (e.g. `stderr.columns` in terminals where stderr is still a TTY, or `COLUMNS` set by the parent), wrapping works exactly as before.

## Changes

| File | Change |
|---|---|
| `src/render/index.ts` | `getTerminalWidth()` returns `null` instead of `UNKNOWN_TERMINAL_WIDTH`; wrap pass conditioned on `terminalWidth !== null`; `canCombine` logic updated to not merge lines when width is unknown |
| `tests/render-width.test.js` | Updated fallback test to verify lines are NOT truncated when width is unavailable |

## Test plan

- [x] `npm run build` compiles cleanly
- [x] All 13 render-width tests pass
- [x] Wrapping still works when `stdout.columns` is set (mocked in tests)
- [x] Wrapping still works when `stderr.columns` is set
- [x] Wrapping still works when `COLUMNS` env var is set
- [x] No wrapping occurs when all width sources are unavailable
- [x] Context/Usage lines stay separate in expanded mode when width is unknown
- [x] Manual verification in Claude Code with compact layout

### DIFF
--- a/src/render/index.ts
+++ b/src/render/index.ts
@@ -15,7 +15,6 @@ import {
   renderSessionTokensLine,
 } from './lines/index.js';
 import { dim, RESET } from './colors.js';
-import { UNKNOWN_TERMINAL_WIDTH } from '../utils/terminal.js';
 
 // eslint-disable-next-line no-control-regex
 const ANSI_ESCAPE_PATTERN = /^(?:\x1b\[[0-9;]*m|\x1b\][^\x07\x1b]*(?:\x07|\x1b\\))/;
@@ -25,11 +24,7 @@ const GRAPHEME_SEGMENTER = typeof Intl.Segmenter === 'function'
   ? new Intl.Segmenter(undefined, { granularity: 'grapheme' })
   : null;
 
-function stripAnsi(str: string): string {
-  return str.replace(ANSI_ESCAPE_GLOBAL, '');
-}
-
-function getTerminalWidth(): number {
+function getTerminalWidth(): number | null {
   const stdoutColumns = process.stdout?.columns;
   if (typeof stdoutColumns === 'number' && Number.isFinite(stdoutColumns) && stdoutColumns > 0) {
     return Math.floor(stdoutColumns);
@@ -47,7 +42,10 @@ function getTerminalWidth(): number {
     return envColumns;
   }
 
-  return UNKNOWN_TERMINAL_WIDTH;
+  // No reliable width source available (e.g. piped statusline subprocess).
+  // Return null so callers can skip width-dependent wrapping rather than
+  // using a wrong fallback that causes incorrect line breaks.
+  return null;
 }
 
 function splitAnsiTokens(str: string): Array<{ type: 'ansi' | 'text'; value: string }> {
@@ -143,6 +141,10 @@ function visualLength(str: string): number {
     }
   }
   return width;
+}
+
+function stripAnsi(str: string): string {
+  return str.replace(ANSI_ESCAPE_GLOBAL, '');
 }
 
 function sliceVisible(str: string, maxVisible: number): string {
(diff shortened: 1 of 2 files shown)


crhan:
We hit the same issue in our downstream project (lia HUD integration) and independently arrived at an equivalent fix.

**Root cause**: `getTerminalWidth()` falls back to `UNKNOWN_TERMINAL_WIDTH` (40) in subprocess/piped contexts. Both `canCombine` in `renderExpanded()` and `wrapLineToWidth` in `render()` treat 40 as a real width, causing:
1. Context + Usage lines are never combined (40 chars is too narrow)
2. Already-rendered lines get re-wrapped at 40 chars

**Our local patch** (validated and running in production):
```ts
// canCombine fix
const widthIsReal = terminalWidth && terminalWidth !== UNKNOWN_TERMINAL_WIDTH;
const canCombine = !widthIsReal || visualLength(combinedLine) <= terminalWidth;

// wrapLineToWidth fix
const wrapWidth = (terminalWidth && terminalWidth !== UNKNOWN_TERMINAL_WIDTH) ? terminalWidth : 0;
const visibleLines = physicalLines.flatMap(line => wrapLineToWidth(line, wrapWidth));
```

Your approach of returning `null` from `getTerminalWidth()` is cleaner — it eliminates the sentinel constant entirely. +1 for this PR.

Just flagging: make sure the `wrapLineToWidth` call path is also covered (same root cause, easy to miss).

TravelThePath:
Thanks for the flag! Both paths are covered in this PR:

1. **`wrapLineToWidth`** (`render()` L494-496) — guarded by `terminalWidth ?`, skips wrapping entirely when `null`
2. **`canCombine`** (`renderExpanded()` L402) — changed from `!terminalWidth` to `terminalWidth !== null`, so lines combine freely when width is unknown

Same root cause, both addressed.

jarrodwatts:

PR #389, opened 2026-04-05
fix: prefer COLUMNS over misreported stdout width

### LINKED ISSUE #385: [Bug] Windows: terminal width detected as ~50% of actual width, add            terminalWidth config override

Gujiassh:
## Summary
- treat the `COLUMNS` env var as a hard terminal-width override before `process.stdout.columns`
- keep Windows Claude hook subprocesses from rendering the HUD at roughly half width when stdout columns are misreported
- add focused terminal-width regression coverage and update the published `dist/` output

Fixes #385.

## Testing
- `npm run build`
- `node --test dist/utils/terminal.test.js`


Gujiassh:
I tightened this branch's test coverage by moving the new terminal assertions into the runtime test suite that `npm test` actually exercises, and removing the extra source-level test file that Node was picking up directly.

Local verification:
- `npm test` now passes the branch-local terminal assertions
- the only remaining failure in the full suite is `countConfigs cache: miss on nested rules additions`, which is outside this PR's terminal-width scope

So the COLUMNS/stdout-width behavior on this branch is covered and locally green; the remaining red is an unrelated baseline failure.

Gujiassh:
Follow-up: I pushed a small branch-local test fix so the new terminal-width assertions run through the same runtime test suite as the rest of the project, instead of being picked up as a separate source-level Node test.

Local verification after the update:
- the COLUMNS/stdout-width assertions are green inside `npm test`
- the remaining full-suite failure is still `countConfigs cache: miss on nested rules additions`, which is outside this PR's terminal-width scope

So this branch's terminal behavior is now covered and locally verified; the remaining red appears to be unrelated baseline noise.

Gujiassh:
Refresh check on current `fix/windows-columns-override` head `975fa14` (already rebased onto latest `origin/main` `bfde7cf`, no behind commits).

Revalidated locally:
- `npm run build`
- `node --test tests/terminal.test.js` (13 passed)

No additional code changes were needed in this refresh pass.

jarrodwatts:

PR #390, opened 2026-04-06
fix(setup): use sort -V for Windows Git Bash to avoid awk quoting breakage

### LINKED ISSUE #326: claude-hud StatusLine Not Working via Git Bash on Windows
# claude-hud StatusLine Not Working via Git Bash on Windows

  ### Summary

  On Windows, when Claude Code is launched via Git Bash, the `statusLine` command configured for claude-hud fails to execute. The HUD does not appear. However, the same Claude Code instance launched via PowerShell works correctly with HUD displayed.

  ### Steps to Reproduce

  1. Windows 10/11 with Git Bash (Git for Windows) installed
  2. Launch Claude Code by running `claude` in Git Bash
  3. claude-hud statusLine is configured in `settings.json`
  4. No HUD appears below the input field

  ### Expected Behavior

  HUD should display below the input field, showing model name, context usage, and other configured information.

  ### Actual Behavior

  HUD is not displayed at all. Even with a minimal test command like `bash -c 'echo TEST_STATUSLINE'`, nothing appears. Manual testing of the same command in Git Bash works correctly, confirming the command itself is valid.

  ### Environment

  - OS: Windows 10 Pro 10.0.19045
  - Node/Bun version: Node v22.14.0
  - Claude Code version: v2.1.84

  ### Configuration

  ```json
  "statusLine": {
  "type": "command",
  "command": "bash -c '\"/c/Program Files/nodejs/node\" \"$(ls -td ~/.claude/plugins/cache/claude-hud/claude-hud/*/ 2>/dev/null | head -1)dist/index.js\"' 2>&1"
  }
```
 Tested bash command (works in Git Bash terminal, but not via Claude Code):
  ``bash -c '"/c/Program Files/nodejs/node" "$(ls -td ~/.claude/plugins/cache/claude-hud/claude-hud/*/ 2>/dev/null | head -1)dist/index.js"' 2>&1``
  When executed directly in Git Bash terminal, this outputs:
  [claude-hud] Initializing...
  However, when configured in settings.json and launched via Git Bash, the HUD still does not appear.

 ### Logs or Screenshots

  Debugging confirms:
  - which node in Git Bash returns /c/Program Files/nodejs/node
  - Manual execution of the statusLine command in Git Bash produces correct output
  - The same command configured in statusLine does not execute when Claude Code is launched via Git Bash
  - The same configuration works when Claude Code is launched via PowerShell

### PR DESCRIPTION
**Pull request by TixXin**
## Summary

- The macOS/Linux awk-based version sorting command uses `'"'"'` to nest single quotes inside `bash -c '...'`. After JSON encoding/decoding in `settings.json`, this quoting breaks on Windows Git Bash, causing a silent syntax error that prevents the HUD statusline from starting.
- Provides a dedicated Windows + Git Bash command template using `sort -V` (GNU version sort, included with Git for Windows) instead of reusing the macOS/Linux awk-based format, avoiding nested single-quote issues entirely.

## Root Cause

In `setup.md` Step 1, the "Windows + Git Bash" section previously said "Use the macOS/Linux bash command format above". That format contains:

```
awk -F/ '"'"'{ print $(NF-1) "\t" $(0) }'"'"'
```

When this command is written into `settings.json` and later decoded by Claude Code, the `'"'"'` quoting sequence breaks under Windows Git Bash, producing a syntax error that silently prevents the HUD process from starting.

## Fix

Replace the awk pipeline with `sort -V` for the Windows + Git Bash section only:

```bash
# Before (broken on Windows Git Bash after JSON round-trip):
ls -d ... | awk -F/ '"'"'{ print $(NF-1) "\t" $(0) }'"'"' | sort -t. -k1,1n ... | tail -1 | cut -f2-

# After (works on Windows Git Bash):
ls -1d ... | sort -V | tail -1
```

macOS/Linux commands are unchanged since they don't have this quoting issue.

## Test plan

- [x] Verified `sort -V` correctly resolves the latest plugin version on Windows Git Bash
- [x] Verified the generated command successfully starts the HUD process
- [ ] Verify macOS/Linux commands remain unaffected (no changes to those sections)

Fixes #326

### DIFF
--- a/commands/setup.md
+++ b/commands/setup.md
@@ -175,16 +175,16 @@ On Windows prefer `node` first and only fall back to `bun`.
 
 **Important**: Do **not** reuse the macOS/Linux awk-based command on Windows + Git Bash. The `awk` fragment requires `'"'"'` quoting to nest single quotes inside `bash -c '...'`. After JSON encoding and decoding, this quoting breaks on Windows Git Bash, causing a silent syntax error that prevents the HUD process from starting (see [#326](https://github.com/jarrodwatts/claude-hud/issues/326)).
 
-Instead, use `sort -V` (GNU version sort, included with Git for Windows) which avoids nested single quotes entirely:
+Instead, use `sort -V` (GNU version sort, included with Git for Windows) which avoids nested single quotes entirely. The command still exports `COLUMNS` so the HUD receives the real terminal width, and it uses the marketplace-aware cache glob:
 
    **When runtime is bun** - add `--env-file /dev/null` to prevent Bun from auto-loading project `.env` files:
    ```
-   bash -c 'plugin_dir=$(ls -1d "${CLAUDE_CONFIG_DIR:-$HOME/.claude}/plugins/cache/claude-hud/claude-hud"/*/ 2>/dev/null | sort -V | tail -1); exec "{RUNTIME_PATH}" --env-file /dev/null "${plugin_dir}{SOURCE}"'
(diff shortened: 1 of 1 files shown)


jarrodwatts (approved):
Rebased this onto the current setup templates and kept the Windows Git Bash sort -V simplification without regressing the newer COLUMNS export or marketplace-aware cache path lookup.

PR #391, opened 2026-04-06
fix: restore null fallback for unknown terminal width

nikolay-bratanov:
## Summary

- `UNKNOWN_TERMINAL_WIDTH` was changed from `null` (v0.0.11) to `40` in v0.0.12, causing HUD output to be aggressively wrapped/truncated to 40 columns when running as a statusline subprocess (where `stdout.columns` and `stderr.columns` are unavailable)
- Restore `null` fallback and add `null` guards in `truncateToWidth` and `wrapLineToWidth` so wrapping is skipped when the real terminal width cannot be detected

## Test plan

- [x] Verified HUD renders full-width output after the fix (macOS, Claude Code statusline)
- [ ] Confirm no regression on terminals where width _is_ available (wrapping should still work normally)

### DIFF
--- a/dist/render/index.js
+++ b/dist/render/index.js
@@ -154,7 +154,7 @@ function sliceVisible(str, maxVisible) {
     return result;
 }
 function truncateToWidth(str, maxWidth) {
-    if (maxWidth <= 0 || visualLength(str) <= maxWidth) {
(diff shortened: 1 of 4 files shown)


jarrodwatts:

PR #393, opened 2026-04-07
fix: honor COLUMNS override for HUD width

### LINKED ISSUE #385: [Bug] Windows: terminal width detected as ~50% of actual width, add            terminalWidth config override

voidborne-d:
## Summary
- treat `COLUMNS` as a hard override before probing stream widths so Windows hook subprocesses can force the real terminal width
- reuse the same terminal-width resolution path for both rendering and adaptive progress bars
- add regression coverage for env override precedence plus stderr fallback behavior

## Testing
- `npm test`

Fixes #385


jarrodwatts:

PR #397, opened 2026-04-07
fix: increase UNKNOWN_TERMINAL_WIDTH fallback from 40 to 120

Rektoooooo:
## Summary

- When Claude Code runs the statusline as a subprocess, `process.stdout.columns` is `undefined` (piped, no TTY). The plugin falls back to `UNKNOWN_TERMINAL_WIDTH` which was **40 columns**
- This causes aggressive line wrapping: the project line (model + path + git + session name + version + duration + cost) splits into 4+ separate lines, and context + usage can never combine — even on a 150+ column terminal
- Bumps the fallback to **120** which matches a typical terminal width. Claude Code's own rendering handles any overflow, so over-estimating is safe while under-estimating breaks the layout

### Before (40 col fallback)
```
[Opus 4.6 (1M context)]
WorkSebas/Downloads
precious-sauteeing-hollerith
CC v2.1.92 | ⏱ 30m
Context ████ 11%
Usage ████ 22% (resets in 4h 3m)
```

### After (120 col fallback)
```
[Opus 4.6 (1M context)] | Downloads/claude-hud git:(main* [+5 -5]) | precious-sauteeing-hollerith | CC v2.1.92 | ⏱ 35m
Context ████ 12% | Usage ████ 24% (resets in 3h 58m)
```

## Test plan
- [x] Verified locally by patching the installed plugin and observing the HUD immediately combining lines
- [x] Run `npm run build && npm test` to confirm no regressions


jarrodwatts:

PR #406, opened 2026-04-08
fix(transcript): show 'agent' instead of 'unknown' for general-purpose agents

### LINKED ISSUE #405: fix: show 'agent' instead of 'unknown' for general-purpose agents
## Problem

When the `Agent` tool is invoked without a `subagent_type` (i.e. using the default general-purpose agent), the transcript's `input` block contains no `subagent_type` field. The current fallback in `src/transcript.ts` is `'unknown'`, which appears in the agents line as:

```
✓ unknown: Research some topic (4m 22s)
```

This is confusing — the agent ran successfully, it just doesn't have a named type.

## Root cause

In `src/transcript.ts`:

```ts
type: (input?.subagent_type as string) ?? 'unknown',
```

When `subagent_type` is absent (general-purpose agent), the label becomes `unknown`.

## Fix

Change the fallback from `'unknown'` to `'agent'`:

```ts
type: (input?.subagent_type as string) ?? 'agent',
```

This makes the display more intuitive:

```
✓ agent: Research some topic (4m 22s)
```

A PR with this one-line fix is ready.

Hudson7777:
## Summary

- Changes the fallback agent type label from `'unknown'` to `'agent'` when `subagent_type` is absent in the transcript

## Problem

When the `Agent` tool is invoked without a `subagent_type` (default general-purpose agent), the agents line displays:

```
✓ unknown: Research some topic (4m 22s)
```

The label `unknown` is misleading — the agent ran fine, it just has no named type.

## Fix

One-line change in `src/transcript.ts`:

```diff
- type: (input?.subagent_type as string) ?? 'unknown',
+ type: (input?.subagent_type as string) ?? 'agent',
```

Result:

```
✓ agent: Research some topic (4m 22s)
```

## Testing

- [x] `npm run build` passes with no errors
- [x] Verified locally: general-purpose agents now show `agent` instead of `unknown`

Closes #405

### DIFF
--- a/src/transcript.ts
+++ b/src/transcript.ts
@@ -296,7 +296,7 @@ function processEntry(
         const input = block.input as Record<string, unknown>;
         const agentEntry: AgentEntry = {
           id: block.id,
-          type: (input?.subagent_type as string) ?? 'unknown',
+          type: (input?.subagent_type as string) ?? 'agent',
           model: (input?.model as string) ?? undefined,
           description: (input?.description as string) ?? undefined,
           status: 'running',


jarrodwatts:
Closing as superseded by #456, which is now merged. I kept the same fix direction there but patched the final label to the shorter maintainer-preferred .

PR #409, opened 2026-04-08
fix: detect real terminal width via ancestor TTY process tree walk

Rektoooooo:
## Summary

- When Claude Code runs the statusline as a subprocess, both `stdout` and `stderr` are piped — so `process.stdout.columns`, `process.stderr.columns`, and `process.env.COLUMNS` are all unavailable
- Adds `detectTtyWidth()` in `src/utils/terminal.ts` that walks up the process tree via `ps` to find an ancestor with a real TTY, then reads its column count with `stty size`
- The result is cached so the tree walk only happens once per invocation
- Falls back to `UNKNOWN_TERMINAL_WIDTH` (120) only if no TTY ancestor is found
- This supersedes #397 — it includes the 120 fallback bump AND adds real detection

### How it works

```
Process tree example:
  PID 7321 (node/hud)     TTY: ?? ← stdout piped, no TTY
  PID 7319 (bash)          TTY: ?? ← also piped
  PID 5253 (zsh)           TTY: ttys000 ← found it!
    → stty size < /dev/ttys000 → "51 167"
    → terminal width = 167 ✓
```

### Before (static 40 col fallback)
```
[Opus 4.6 (1M context)]
WorkSebas/Downloads
precious-sauteeing-hollerith
CC v2.1.92 | ⏱ 30m
Context ████ 11%
Usage ████ 22% (resets in 4h 3m)
```

### After (real width detected — e.g. 167 cols)
```
[Opus 4.6 (1M context)] | Downloads/claude-hud git:(main* [+5 -5]) | precious-sauteeing-hollerith | CC v2.1.92 | ⏱ 35m
Context ████ 12% | Usage ████ 24% (resets in 3h 58m)
```

## Test plan
- [x] Verified locally — `detectTtyWidth()` correctly returns 167 on macOS by finding `/dev/ttys000` via ancestor PID
- [x] Updated `render-width.test.js` fallback test to mock `process.ppid` and reset the TTY cache, ensuring the static fallback path is still tested
- [x] `npm test` passes (1 pre-existing unrelated failure in `loadConfig` elementOrder defaults)
- [x] Cached result avoids repeated `execSync` calls — zero perf impact after first render


jarrodwatts:

PR #410, opened 2026-04-09
fix: use COLUMNS env var and raise default width for compact layout

### LINKED ISSUE #408: lineLayout: compact has no visible effect on display
## Description

Setting `"lineLayout": "compact"` in `~/.claude/plugins/claude-hud/config.json` does not change the display from multi-line to single-line. The HUD continues to render the same way as `"expanded"` mode.

## Steps to Reproduce

1. Install claude-hud plugin and complete setup
2. Edit `~/.claude/plugins/claude-hud/config.json`:
   ```json
   {
     "lineLayout": "compact",
     "display": {
       "showTools": true,
       "showSessionName": true
     }
   }
   ```
3. Interact with Claude Code to trigger a re-render
4. Also tried restarting Claude Code entirely

## Expected Behavior

The HUD should condense into a single line (or fewer lines) as described in the README:
> Setting `"lineLayout": "compact"` produces a single-line condensed view

## Actual Behavior

The display remains identical to `"expanded"` mode — no visible difference.

## Environment

- macOS (darwin, Apple Silicon)
- Claude Code v2.1.96
- claude-hud v0.0.12
- Runtime: bun 1.3.11

## Analysis

Looking at the source code, `renderCompact()` calls `renderSessionLine()` which joins all parts with ` | ` into one line, while `renderExpanded()` renders each element on its own line. The config is read from disk on every invocation via `loadConfig()`, so caching shouldn't be the issue. 

It's possible that Claude Code's statusLine rendering treats each `console.log()` call or `\n` as a separate line and the compact output is still being split by the terminal or by Claude Code's status line handler.

kagura-agent:
## Problem

When Claude Code runs the statusLine plugin as a subprocess, `stdout` is piped and `process.stderr.columns` returns 0. The fallback `UNKNOWN_TERMINAL_WIDTH = 40` kicks in, causing even a compact one-liner to wrap into 5+ lines at 40 chars — making `compact` mode indistinguishable from `expanded`.

Community workaround: `export COLUMNS` in `.zshrc` (confirmed working by multiple users in #408).

## Fix

1. **Raise `UNKNOWN_TERMINAL_WIDTH` from 40 to 120** — a reasonable modern terminal default. This fixes the issue for users who don't have `COLUMNS` exported.

2. **Inject `export COLUMNS` in setup command** — the bash wrapper in `setup.md` now exports `COLUMNS` (using `tput cols` with 120 fallback) before `exec`ing the runtime, so the subprocess inherits the real terminal width via `process.env.COLUMNS`.

The existing `getTerminalWidth()` already reads `process.env.COLUMNS` as a fallback, so no changes needed there.

## Testing

- All 352 tests pass
- Updated the fallback-width test assertion to match the new default

Closes #408


kagura-agent:
Closing — reducing PR volume. Happy to reopen if useful.

PR #413, opened 2026-04-09
fix: fallback to tput cols for terminal width detection in subprocess

fyeeme:
## Summary

- Terminal width detection fails when bun/node runs as a statusline subprocess: both `process.stdout.columns` and `process.stderr.columns` are `undefined` (stdout is piped, stderr may also lack TTY info)
- The existing fallback chain (`stdout → stderr → $COLUMNS → hardcoded 40`) returns 40 columns, causing aggressive truncation even on wide terminals
- Added `tput cols` as an additional fallback before the hardcoded 40 — `tput` queries the terminfo database and correctly returns the terminal width regardless of TTY fd availability

## Problem

When claude-hud runs as a statusline subprocess via `bash -c '... bun "${plugin_dir}src/index.ts"'`:
- `process.stdout.columns` → `undefined` (stdout is a pipe)
- `process.stderr.columns` → `undefined` (no TTY in subprocess)
- `$COLUMNS` → not set
- Falls back to `UNKNOWN_TERMINAL_WIDTH = 40` → truncates everything beyond 40 chars

This means git branch names like `feat/msg_v3_0506_v2_0306` get truncated to `feat/msg_v3_0506_v...` even on a 200-column terminal.

## Fix

Added a `tput cols` fallback via `execSync` (with 1s timeout) between the `$COLUMNS` check and the hardcoded 40 default. Verified that `tput cols` returns the correct width even in nested `bash -c → bun` subprocess chains.

| Metric | Before | After |
|--------|--------|-------|
| Detected width (80-col terminal) | 40 | 80 |
| Git branch display | `feat/msg_v3_05...` | `feat/msg_v3_0506_v2_0306` |
| Overhead | 0 | ~5ms (`tput cols`) |

## Test plan

- [x] Verified `tput cols` returns correct width in nested `bash -c → bun` subprocess
- [x] Verified truncation is resolved with all HUD elements enabled on compact layout
- [ ] Tested on Linux (where `tput` is also available)
- [ ] Confirmed no regression when `tput` is unavailable (graceful fallback to 40)

### DIFF
--- a/src/render/index.ts
+++ b/src/render/index.ts
@@ -16,6 +16,7 @@ import {
 } from './lines/index.js';
 import { dim, RESET } from './colors.js';
 import { UNKNOWN_TERMINAL_WIDTH } from '../utils/terminal.js';
+import { execSync } from 'child_process';
 
 // eslint-disable-next-line no-control-regex
 const ANSI_ESCAPE_PATTERN = /^(?:\x1b\[[0-9;]*m|\x1b\][^\x07\x1b]*(?:\x07|\x1b\\))/;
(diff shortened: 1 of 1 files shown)


jarrodwatts:

Issue #404, opened 2026-04-08
Statusline wraps to multiple rows when stdout/stderr columns unavailable


pfe-nazaries:
## Problem

When claude-hud runs as a statusline subprocess, `process.stdout.columns` and `process.stderr.columns` are both unavailable (pipes, not TTYs). The fallback in `src/utils/terminal.ts` is `UNKNOWN_TERMINAL_WIDTH = 40`, and `getTerminalWidth()` in `src/render/index.ts` also falls back to 40.

40 columns is too narrow for most HUD content, causing `wrapLineToWidth` to break lines and the HUD displays on multiple rows instead of a single compact line.

## Current behavior

```
src/utils/terminal.ts:1  →  UNKNOWN_TERMINAL_WIDTH = 40
src/render/index.ts:50   →  return UNKNOWN_TERMINAL_WIDTH;  // fallback
```

Lines wider than 40 chars get wrapped via `wrapLineToWidth()` at line 488.

## Workaround

Setting `COLUMNS=200` in the statusline command:

```json
"statusLine": {
  "type": "command",
  "command": "bash -c 'COLUMNS=200 bun .../src/index.ts'"
}
```

This works because `getTerminalWidth()` checks `process.env.COLUMNS` before the 40-char fallback.

## Suggested fix

Either:
1. **Increase `UNKNOWN_TERMINAL_WIDTH`** to something reasonable (e.g. 120 or 200) since the 40-char case is almost never correct for a modern terminal.
2. **Add a `terminalWidth` config option** in `config.json` so users can set it without modifying the command.
3. **Try `tput cols`** as an additional fallback before defaulting to 40.

## Environment

- macOS (Darwin 24.5.0)
- claude-hud 0.0.12
- Running as statusline subprocess via Claude Code CLI

Hudson7777:
## Additional reproduction: iTerm2 on macOS

Reproduced this on **iTerm2** (macOS Darwin 24.6.0, claude-hud 0.0.12, Claude Code 2.1.96).

**Environment details:**
- Terminal: iTerm2, actual window width = 272 columns
- `process.stdout.columns` → unavailable (piped)
- `process.stderr.columns` → unavailable
- `process.env.COLUMNS` → `0` (Claude Code does not propagate `COLUMNS` to the statusline subprocess)
- `tput cols` inside the subprocess → returns `80` (the inherited TTY default at launch time, not the actual window width)

[...]

PR #414, opened 2026-04-09
fix: keep compact status lines from wrapping when terminal width is unknown

voidborne-d:
## Summary\n- raise the unknown terminal width fallback from 40 to 120 for piped statusline runs\n- add a regression test that simulates stdout/stderr column metadata being unavailable\n- keep compact HUD output on one line in the common statusline subprocess case\n\n## Testing\n- npm test -- --test-name-pattern='fallback width|stdout columns over COLUMNS env fallback'\n\nFixes #404


jarrodwatts:

PR #415, opened 2026-04-10
feat: make usage reset label configurable

### LINKED ISSUE #411: Feature request: make "resets in" label in usage line configurable
## Description

In expanded layout, the usage line displays something like:

```
19% (重置剩余 3h 17m)  |  本周  6% (重置剩余 5d 21h)
```

The "resets in" / "重置剩余" label before the countdown feels redundant — the time value alone already conveys the meaning clearly. Removing it would also help shorten the line length, reducing the chance of line wrapping on narrower terminals.

## Proposal

Add a config option (e.g. `display.showResetLabel: false`) to hide the "resets in" prefix, so the output becomes:

```
19% (3h 17m)  |  本周  6% (5d 21h)
```

The label text is defined in `src/i18n/{en,zh}.ts` as `format.resetsIn` and referenced in `src/render/lines/usage.ts` and `src/render/session-line.ts`.

zviehrman13:
## Summary

Adds a `display.showResetLabel` config option so users can hide the `resets in` prefix from usage countdowns.

Examples:

- Before: `Usage 5h 19% (resets in 3h 17m)`
- After: `Usage 5h 19% (3h 17m)`

This applies to both:
- expanded usage rendering
- compact/session-line rendering

## Why

The reset prefix is a bit redundant, and removing it shortens the HUD line, which helps reduce clutter and wrapping risk.

## Changes

- add `display.showResetLabel` to config, default `true`
- respect it in `src/render/lines/usage.ts`
- respect it in `src/render/session-line.ts`
- document it in `README.md`
- add tests for text and bar modes

## Testing

- [x] `npm test`
- [x] verified real plugin output locally before/after using stdin-driven render

Closes #411

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,8 @@ Chinese HUD labels are available as an explicit opt-in. English stays the defaul
 | `display.showUsage` | boolean | true | Show Claude subscriber usage limits when available |
 | `display.usageBarEnabled` | boolean | true | Display usage as visual bar instead of text |
 | `display.timeFormat` | `relative` \| `absolute` \| `both` | `relative` | How reset times are shown in usage windows: countdown only (`resets in 2h 30m`), wall-clock time (`resets at 14:30`), or both (`resets in 2h 30m, at 14:30`) |
+| `display.showResetLabel` | boolean | true | Show the `resets in` prefix before usage countdowns |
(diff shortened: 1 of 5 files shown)


jarrodwatts (approved):
The toggle stays opt-in, now composes correctly with relative/absolute/both reset formatting, and the rebased branch passed the full test suite on current main. Merging.

PR #418, opened 2026-04-10
render -- instead of 0% before context is initialized

### LINKED ISSUE #417: Initial context should not be 0%
**Title:** `/context` shows 0% usage but breakdown reports 18.2k tokens already consumed

**Environment:**
- Model: Opus 4.6 (`claude-opus-4-6`)
- Fresh session, no user input

**Bug:**
On a brand-new session with nothing sent, the HUD shows `Context ░░░░░░░░░░ 0%`, but running `/context` reports **18.2k/200k tokens (9%)** already used:

- System prompt: 6.4k (3.2%)
- System tools: 8.6k (4.3%)
- Memory files: 2.5k (1.2%)
- Skills: 598 (0.3%)
- Messages: 110 (0.1%)
- Autocompact buffer: 33k (16.5%)

**Expected:** The HUD's 0% and `/context`'s 9% should agree. Either the HUD is under-reporting, or `/context` is counting overhead that shouldn't be shown to the user.

**Impact:** Users can't trust the context indicator to know how much room they actually have.

horno1337:
## Summary

- Before the first API call, Claude Code reports `null` usage — showing `0%` is misleading since the context window already has overhead (system prompt, tools, memory files, etc.)
- Adds `isContextInitialized()` to detect uninitialized state
- Renders `--` instead of `0%` until real data arrives

Fixes #417

## Test plan

- [ ] Start a new session and verify the context bar shows `--` before the first response
- [ ] Confirm it switches to a real percentage after the first API call
- [ ] Run existing tests: `npm test`

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -100,6 +100,7 @@ Lines 1-2 always shown. Additional lines are opt-in via config:
 
 | Threshold | Color | Action |
 |-----------|-------|--------|
+| uninitialized | — | Show `--` (no API call yet) |
 | <70% | Green | Normal |
 | 70-85% | Yellow | Warning |
 | >85% | Red | Show token breakdown |
--- a/src/render/lines/identity.ts
+++ b/src/render/lines/identity.ts
@@ -3,6 +3,7 @@ import {
   getContextPercent,
   getBufferedPercent,
   getTotalTokens,
+  isContextInitialized,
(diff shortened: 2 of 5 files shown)


jarrodwatts:
Superseded by #430, which fixes the underlying zero-percent fallback in the shared parsing path.

PR #420, opened 2026-04-11
feat(config): add maxWidth fallback for terminal width detection

### LINKED ISSUE #385: [Bug] Windows: terminal width detected as ~50% of actual width, add            terminalWidth config override

### LINKED ISSUE #404: Statusline wraps to multiple rows when stdout/stderr columns unavailable

macworld:
## Summary

Adds a `maxWidth` config option (`number | null`, default `null`) that replaces the hardcoded `UNKNOWN_TERMINAL_WIDTH` fallback when terminal width detection fails.

- When the terminal reports a real width (via `stdout.columns`, `stderr.columns`, or `COLUMNS` env), that width is used — `maxWidth` is **ignored**
- Only when all detection fails and the code falls back to `UNKNOWN_TERMINAL_WIDTH` (40), `maxWidth` is used instead
- When `maxWidth` is not set, the existing 40-column fallback is preserved

```json
{
  "maxWidth": 50
}
```

### Motivation

Several users have reported terminal width issues:

- **#385**: Windows Terminal detects ~50% of actual width, needs a way to tune the fallback
- **#404**: When stdout/stderr columns are unavailable (piped subprocesses, mobile clients), the 40-column fallback is too narrow

The hardcoded `UNKNOWN_TERMINAL_WIDTH = 40` is a reasonable default but not ideal for all devices. `maxWidth` lets users tune it without affecting environments where auto-detection works correctly.

### Priority logic

```
Terminal width detected (e.g. 120)  →  use 120, maxWidth ignored
Detection fails + maxWidth set      →  use maxWidth
Detection fails + no maxWidth       →  use UNKNOWN_TERMINAL_WIDTH (40)
```

### Changes

| File | Change |
|------|--------|
| `src/config.ts` | Add `maxWidth: number \| null` to `HudConfig`, validation in `mergeConfig()` |
| `src/render/index.ts` | Use `maxWidth` as fallback only when `getTerminalWidth()` returns `UNKNOWN_TERMINAL_WIDTH` |
| `tests/config.test.js` | Validation: default null, valid values, invalid values (0, negative, NaN, Infinity, string) |
| `tests/render-width.test.js` | Two render tests: maxWidth used as fallback when detection fails; maxWidth ignored when terminal width is detected |

## Testing

- [x] `npm test` — all tests pass (1 pre-existing flaky failure in `core.test.js` unrelated to this change)
- [x] `npm run build` — compiles cleanly

## Checklist

- [x] Docs updated if behavior changed — user-facing config option is self-documenting; README update can follow separately

Closes #385, closes #404

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ After choosing a preset, you can turn individual elements on or off.
 ### Manual Configuration
 
 Edit `~/.claude/plugins/claude-hud/config.json` directly for advanced settings such as `colors.*`,
-`pathLevels`, threshold overrides, and `display.timeFormat`. Running `/claude-hud:configure`
+`pathLevels`, `maxWidth`, threshold overrides, and `display.timeFormat`. Running `/claude-hud:configure`
 preserves those manual settings while still letting you change `language`, layout, and the common
 guided toggles.
 
@@ -159,6 +159,7 @@ Chinese HUD labels are available as an explicit opt-in. English stays the defaul
 | `language` | `en` \| `zh` | `en` | HUD label language. English is the default; set `zh` to enable Chinese labels. |
 | `lineLayout` | string | `expanded` | Layout: `expanded` (multi-line) or `compact` (single line) |
 | `pathLevels` | 1-3 | 1 | Directory levels to show in project path |
+| `maxWidth` | number \| `null` | `null` | Optional fallback width used only when terminal width detection fails completely |
 | `elementOrder` | string[] | `["project","context","usage","memory","environment","tools","agents","todos"]` | Expanded-mode element order. Omit entries to hide them in expanded mode. |
 | `gitStatus.enabled` | boolean | true | Show git branch in HUD |
 | `gitStatus.showDirty` | boolean | true | Show `*` for uncommitted changes |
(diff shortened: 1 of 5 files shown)


jarrodwatts (approved):
The fallback stays opt-in, keeps real terminal detection as the primary path, and passed the full test suite on current main. Added the missing README docs for discoverability. Merging.

PR #422, opened 2026-04-11
fix: detect Windows terminal width via PowerShell when ConPTY reports 0

danielyang1009:
## Problem

  On Windows, `process.stdout.columns` and `process.stderr.columns` both
  return `0` when Claude Code runs the statusLine command as a subprocess
  with piped stdout (ConPTY architecture). The `COLUMNS` environment
  variable is also unset. The fallback chain in both `getTerminalWidth()`
  and `getAdaptiveBarWidth()` therefore bottoms out at
  `UNKNOWN_TERMINAL_WIDTH = 40`, causing the HUD to wrap into 4+ lines
  instead of the intended 2-line layout.

## Root cause

  Windows Terminal uses ConPTY as an intermediary layer. When Claude Code
  spawns the statusLine subprocess with piped stdout, the inner pipe
  reports `columns = 0` — the actual window width is not propagated
  through the ConPTY layer to the child process's stream properties.

 ## Fix

  Add `getWindowsTerminalWidth()` in `src/utils/terminal.ts`. It queries
  the real terminal width via PowerShell's
  `$Host.UI.RawUI.WindowSize.Width`, which reads from the Windows Console
  API and correctly reflects the visible window width.

  Both `getAdaptiveBarWidth()` (terminal.ts) and `getTerminalWidth()`
  (render/index.ts) now call this as a last-resort fallback before
  returning the unknown-width sentinel.

 ## Tested on

  - Windows 11 + Windows Terminal + Claude Code statusLine
  - Actual terminal width: 120 columns
  - Before fix: HUD rendered as 4 lines
  - After fix: correct 2-line expanded layout

### DIFF
--- a/src/render/index.ts
+++ b/src/render/index.ts
@@ -15,7 +15,7 @@ import {
   renderSessionTokensLine,
 } from './lines/index.js';
 import { dim, RESET } from './colors.js';
-import { UNKNOWN_TERMINAL_WIDTH } from '../utils/terminal.js';
+import { UNKNOWN_TERMINAL_WIDTH, getWindowsTerminalWidth } from '../utils/terminal.js';
 
 // eslint-disable-next-line no-control-regex
 const ANSI_ESCAPE_PATTERN = /^(?:\x1b\[[0-9;]*m|\x1b\][^\x07\x1b]*(?:\x07|\x1b\\))/;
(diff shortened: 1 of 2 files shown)


jarrodwatts:

PR #423, opened 2026-04-11
fix: detect terminal width via tput when stdout/stderr are piped

bxyb:
## Problem

When Claude Code invokes the statusline command, it pipes **both stdout and stderr**, causing `process.stdout.columns` and `process.stderr.columns` to be `undefined`. The existing code comment says stderr should still be connected to the real terminal, but in practice Claude Code redirects it too.

This means `getTerminalWidth()` always falls through to `UNKNOWN_TERMINAL_WIDTH` (40 chars), causing the HUD to wrap every segment onto its own line even on wide terminals — producing 4–5 lines of output instead of 1–2.

Verified by logging inside the subprocess:
```json
{
  "stdout_columns": undefined,
  "stderr_columns": undefined,
  "env_COLUMNS": undefined
}
```

## Fix

Add a `tput cols` fallback as a last resort. `tput` reads the terminal size from `/dev/tty` directly, which works even when all stdio file descriptors are redirected/piped.

Detection chain becomes:
```
stdout.columns → stderr.columns → COLUMNS env → tput cols → fallback (40)
```

The `tput` call has a 50 ms timeout and is fully wrapped in try/catch, so it degrades gracefully on systems where `tput` is unavailable (Windows, minimal containers, etc.).

## Test

```bash
# Simulate Claude Code's piped invocation
echo '<stdin-json>' | node dist/index.js
# Before: wraps at 40 chars — every segment on its own line
# After:  wraps at actual terminal width (e.g. 160) — fits on 1–2 lines
```


jarrodwatts:

PR #424, opened 2026-04-11
fix(setup): inject COLUMNS width fallback into bash statusLine command

### LINKED ISSUE #416: Setup script should inject COLUMNS=$(tput cols) into statusLine command
## Problem

The setup skill (`commands/setup.md`) generates a `statusLine.command` that does **not** set the `COLUMNS` environment variable before launching the HUD process. When running inside terminal multiplexers like **zellij** (and potentially tmux), the subprocess cannot detect the pane width:

- `process.stdout.columns` → unavailable (piped)
- `process.stderr.columns` → unavailable (zellij redirects stderr)
- `process.env.COLUMNS` → `0` (zellij sets this to 0 for subprocesses)

All three fallbacks in `getTerminalWidth()` fail, returning `null` (v0.0.11) or `UNKNOWN_TERMINAL_WIDTH` (v0.0.12). This causes the HUD content to render much narrower than the actual pane, leaving significant empty space on the right side.

**Related**: #404 covers the same root cause from the wrapping/line-split angle. This issue specifically addresses the **setup script** as the fix point.

## Key observation

In zellij (and likely tmux), `tput cols` **correctly returns the pane width**, unlike some other environments where it returns a stale value. This makes it a viable injection point.

```bash
# Inside a zellij pane (actual width ~100 cols):
$ echo $COLUMNS
0
$ tput cols
100          # ← correct pane width
```

## Current generated command (Node variant)

```bash
bash -c 'plugin_dir=$(ls -d ...); exec "node" "${plugin_dir}dist/index.js"'
```

## Suggested fix

Inject `COLUMNS` via `tput cols` at the start of the generated command:

```bash
bash -c 'export COLUMNS=$(tput cols 2>/dev/null || echo 120); plugin_dir=$(ls -d ...); exec "node" "${plugin_dir}dist/index.js"'
```

This should be added to both the bash and bun command templates in `commands/setup.md` (lines 149-157).

### Why `tput cols` over other approaches

| Approach | Pros | Cons |
|----------|------|------|
| `tput cols` | Works in zellij/tmux, queries actual TTY | May return stale value in some terminals (see #404 comment) |
| Hardcoded value | Always works | Not portable, breaks on resize |
| Config option | User-controlled | Manual maintenance |

`tput cols` is imperfect (as noted in #404), but it's strictly better than the current behavior of not setting `COLUMNS` at all. For environments where `tput cols` also fails, users can still manually override.

## Environment

- Linux (WSL2) with zellij terminal multiplexer
- claude-hud 0.0.11
- Claude Code CLI

## Workaround

Manually add `export COLUMNS=$(tput cols 2>/dev/null || echo 80);` to the `statusLine.command` in `~/.claude/settings.json`. This fix survives plugin updates but gets overwritten if `/claude-hud:setup` is re-run.

Gujiassh:
## Summary
- update /claude-hud:setup bash command templates (bun + node) to export COLUMNS before launching HUD
- use a robust width detection chain: tmux display-message -p '#{pane_width}' || tput cols || 120
- keep existing dynamic plugin-version lookup and runtime invocation unchanged

## Why
- addresses #416 where setup-generated commands can run with COLUMNS=0 or missing width in tmux/zellij subprocesses
- gives getTerminalWidth() a reliable env override without requiring manual user edits

## Testing
- node --test tests/terminal.test.js
- rg -n "tmux display-message -p|tput cols" commands/setup.md

Fixes #416.

### DIFF
--- a/commands/setup.md
+++ b/commands/setup.md
@@ -153,12 +153,12 @@ This is a [Claude Code platform limitation](https://github.com/anthropics/claude
 
    **When runtime is bun** - add `--env-file /dev/null` to prevent Bun from auto-loading project `.env` files:
    ```
-   bash -c 'plugin_dir=$(ls -d "${CLAUDE_CONFIG_DIR:-$HOME/.claude}"/plugins/cache/claude-hud/claude-hud/*/ 2>/dev/null | awk -F/ '"'"'{ print $(NF-1) "\t" $(0) }'"'"' | sort -t. -k1,1n -k2,2n -k3,3n -k4,4n | tail -1 | cut -f2-); exec "{RUNTIME_PATH}" --env-file /dev/null "${plugin_dir}{SOURCE}"'
+   bash -c 'export COLUMNS=$(tmux display-message -p "#{pane_width}" 2>/dev/null || tput cols 2>/dev/null || echo 120); plugin_dir=$(ls -d "${CLAUDE_CONFIG_DIR:-$HOME/.claude}"/plugins/cache/claude-hud/claude-hud/*/ 2>/dev/null | awk -F/ '"'"'{ print $(NF-1) "\t" $(0) }'"'"' | sort -t. -k1,1n -k2,2n -k3,3n -k4,4n | tail -1 | cut -f2-); exec "{RUNTIME_PATH}" --env-file /dev/null "${plugin_dir}{SOURCE}"'
    ```
 
    **When runtime is node**:
    ```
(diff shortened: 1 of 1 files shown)


Gujiassh:
Refresh note:
- No further code changes after PR creation.
- Re-verified targeted behavior with node --test tests/terminal.test.js (pass).
- Scope stays limited to setup command templates in commands/setup.md for injecting COLUMNS width detection fallback chain.

jarrodwatts:

PR #433, opened 2026-04-12
feat: add display.timeFormat config for absolute/relative reset times

### LINKED ISSUE #432: Feature: Show usage reset time as absolute timestamp
## Summary

Currently the HUD displays usage reset times as relative durations only (e.g. `resets in 2h 30m`). It would be useful to optionally display the reset time as an absolute timestamp (e.g. `resets at 14:30` or `resets at Apr 12, 14:30`).

## Motivation

Relative times require mental arithmetic to know *when* something will happen. An absolute time is immediately actionable — especially useful when planning work sessions around rate limit windows.

## Proposed Config Option

Add a `display.timeFormat` option (or similar) to `config.json`:

```json
{
  "display": {
    "timeFormat": "relative"   // "relative" | "absolute" | "both"
  }
}
```

| Value | Example output |
|-------|---------------|
| `"relative"` (default, current behaviour) | `resets in 2h 30m` |
| `"absolute"` | `resets at 14:30` |
| `"both"` | `resets at 14:30 (in 2h 30m)` |

## Implementation Notes

The logic to change lives in two `formatResetTime()` functions:

- `src/render/lines/usage.ts` lines 129–149 (expanded layout)
- `src/render/session-line.ts` lines 328–348 (compact layout)

Both are identical — extracting them to a shared utility first would make the feature cleaner to add. The absolute time could be rendered with `new Date(resetTime).toLocaleTimeString()` (respects the user's locale/timezone automatically).

YASoftwareDev:
## Summary

Closes #432

Adds a \`display.timeFormat\` config option that lets users choose how usage-window reset times appear in the HUD:

| Value | Example output |
|-------|---------------|
| \`"relative"\` (default — existing behaviour) | \`resets in 2h 30m\` |
| \`"absolute"\` | \`resets at 14:30\` / \`resets at Apr 13 14:30\` (different day) |
| \`"both"\` | \`resets in 2h 30m, at 14:30\` |

The absolute format uses \`toLocaleTimeString\` / \`toLocaleDateString\` so it respects the user's system locale and timezone automatically. The \`both\` mode uses a comma separator (not nested parentheses) to avoid doubled brackets when the caller wraps the reset in its own \`(...)\` parenthetical.

## Changes

- **\`src/config.ts\`** — new \`TimeFormatMode\` union type; \`display.timeFormat\` field with default \`'relative'\`; \`validateTimeFormat\` validator; wired into \`mergeConfig\`
- **\`src/render/format-reset-time.ts\`** _(new)_ — shared utility replacing the previously-duplicated \`formatResetTime\` in \`usage.ts\` and \`session-line.ts\`; uses \`t("format.at")\` so the "at" prefix is i18n-aware
- **\`src/i18n/types.ts\` / \`en.ts\` / \`zh.ts\`** — new \`"format.at"\` key (en: \`"at"\`, zh: \`""\` — zh omits the prefix since \`"重置于"\` already encodes the preposition)
- **\`src/render/lines/usage.ts\`** — removed private \`formatResetTime\`; threads \`timeFormat\` through \`formatUsageWindowPart\`; applies correct i18n preposition in both normal and limit-reached paths
- **\`src/render/session-line.ts\`** — same; bar mode preserves the upstream \`(duration / windowLabel)\` pattern for relative mode but switches to the preposition form for absolute/both (avoids incoherent \`at 14:30 / 5h\`)
- **\`tests/format-reset-time.test.js\`** _(new)_ — 18 unit tests covering all three modes, null/past guards, and \`mergeConfig\` validation
- **\`tests/render.test.js\`** — 5 new integration tests; updated one existing assertion that was written against the old (incorrect) limit-reached preposition

## Test plan

- [x] All existing tests pass (\`npm test\`)
- [x] 18 new unit tests for \`formatResetTime\` all pass
- [x] 5 new integration tests for \`renderUsageLine\` with each \`timeFormat\` mode pass
- [x] \`display.timeFormat: "relative"\` (default) produces identical output to before
- [x] \`display.timeFormat: "absolute"\` shows \`at HH:MM\` for same-day resets
- [x] \`display.timeFormat: "absolute"\` shows \`at Mon DD HH:MM\` for next-day resets
- [x] \`display.timeFormat: "both"\` shows \`Xh Ym, at HH:MM\`
- [x] Invalid \`timeFormat\` values fall back to \`"relative"\` silently
- [x] Limit-reached path uses correct preposition for each mode
- [x] zh locale: absolute mode renders \`重置于 14:30\` (no doubled preposition)

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,9 @@ After choosing a preset, you can turn individual elements on or off.
 ### Manual Configuration
 
 Edit `~/.claude/plugins/claude-hud/config.json` directly for advanced settings such as `colors.*`,
-`pathLevels`, and threshold overrides. Running `/claude-hud:configure` preserves those manual settings while still letting you change `language`, layout, and the common guided toggles.
+`pathLevels`, threshold overrides, and `display.timeFormat`. Running `/claude-hud:configure`
+preserves those manual settings while still letting you change `language`, layout, and the common
+guided toggles.
 
 Chinese HUD labels are available as an explicit opt-in. English stays the default unless you choose `中文` in `/claude-hud:configure` or set `language` in config.
 
@@ -172,6 +174,7 @@ Chinese HUD labels are available as an explicit opt-in. English stays the defaul
 | `display.showSpeed` | boolean | false | Show output token speed `out: 42.1 tok/s` |
 | `display.showUsage` | boolean | true | Show Claude subscriber usage limits when available |
 | `display.usageBarEnabled` | boolean | true | Display usage as visual bar instead of text |
+| `display.timeFormat` | `relative` \| `absolute` \| `both` | `relative` | How reset times are shown in usage windows: countdown only (`resets in 2h 30m`), wall-clock time (`resets at 14:30`), or both (`resets in 2h 30m, at 14:30`) |
 | `display.sevenDayThreshold` | 0-100 | 80 | Show 7-day usage when >= threshold (0 = always) |
 | `display.showTokenBreakdown` | boolean | true | Show token details at high context (85%+) |
 | `display.showTools` | boolean | false | Show tools activity line |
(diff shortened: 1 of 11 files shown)


jarrodwatts (approved):
The feature now has user-facing docs, the locale-sensitive test no longer assumes English month names, and the rebased branch passed the full test suite on current main. Merging.

PR #434, opened 2026-04-12
fix: invalidate cached Claude Code version after in-place updates

### LINKED ISSUE #429: Version cache never invalidates when Claude Code updates in-place
## Bug

`getClaudeCodeVersion()` in `src/version.ts` shows a stale Claude Code version after CC auto-updates.

## Root cause

CC updates by creating a **new binary file** (e.g. `~/.local/share/claude/versions/2.1.104`) and re-pointing the `~/.local/bin/claude` symlink. The **old binary is not deleted**.

The cache validation in `getClaudeCodeVersion()` (lines ~158-170) does:

```js
const cachedBinaryInfo = statResolvedBinary(diskCache.binaryPath);
if (cachedBinaryInfo
    && cachedBinaryInfo.path === diskCache.binaryPath
    && cachedBinaryInfo.mtimeMs === diskCache.binaryMtimeMs) {
    // cache hit → return stale version
}
```

Since `diskCache.binaryPath` stores the **resolved real path** of the old binary (e.g. `.../versions/2.1.97`), and that file still exists on disk with the same mtime, the cache always validates as fresh. `resolveClaudeBinaryFromPath()` — which would correctly follow the symlink to the new binary — is never called.

## Reproduction

```
# Before CC update
$ claude --version → 2.1.97
# HUD caches: binaryPath=".../versions/2.1.97", version="2.1.97"

# After CC auto-update
$ claude --version → 2.1.104
$ readlink -f $(which claude) → .../versions/2.1.104
# HUD still shows "CC v2.1.97" because .../versions/2.1.97 still exists
```

## Suggested fix

Also track the **symlink path** and its resolved target. If the symlink now resolves to a different real path than what's cached, invalidate.

```js
// In cache validation, also check that PATH still resolves to the same binary:
const currentBinary = resolveClaudeBinaryFromPath();
if (!currentBinary || currentBinary.path !== diskCache.binaryPath) {
    // symlink changed → invalidate cache
}
```

Or more minimally: stat the symlink itself (`~/.local/bin/claude`) and track its mtime, which changes when the symlink is re-pointed.

## Workaround

Delete the cache file:
```bash
rm ~/.claude/plugins/claude-hud/.claude-code-version-cache.json
```

## Environment

- OS: Linux
- CC install: `~/.local/bin/claude` → `~/.local/share/claude/versions/<ver>`
- claude-hud: 0.0.12 (verified unchanged from 0.0.11)

voidborne-d:
## Summary
- re-check the currently resolved `claude` binary before trusting the persisted version cache
- invalidate the cache when PATH now resolves to a different on-disk binary, even if the old binary still exists unchanged
- add a regression test that simulates Claude Code updating by repointing the binary target without changing the old file mtime

## Testing
- npm test

Closes #429


jarrodwatts (approved):
This is the right fix direction. I patched the branch before merge to normalize the current resolver result through the same realpath/mtime path the cache uses, which fixed a rebased persistence regression in the test suite. Verified with npm test, and stripped the generated dist output back out of the PR.

PR #440, opened 2026-04-15
fix: show usage and cost for enterprise plan model aliases

conquests:
## Problem

Enterprise Claude Code accounts use model IDs like `opusplan`, `sonnetplan`, and `haikuplan` instead of standard model names. With these model IDs, **usage and cost are completely hidden** in the statusline.

**Root cause:** `getProviderLabel()` returns a truthy value (`'Enterprise'`) for these model IDs, and all three display paths check `if (getProviderLabel(stdin))` to suppress output — the same condition used for Bedrock, which genuinely cannot report usage.

## Changes

- **`src/render/lines/usage.ts`** — narrow hide condition from `getProviderLabel()` to `isBedrockModelId()` so enterprise plans are no longer suppressed
- **`src/cost.ts`** — same fix for cost estimation paths; add enterprise alias entries to `ANTHROPIC_MODEL_PRICING` (`opusplan`, `sonnetplan`, `haikuplan`)
- **`src/stdin.ts`** — add `isEnterpriseModelId()` helper, readable label mapping (`opusplan` → `Claude Opus`, etc.), and update `getProviderLabel()` to return `'Enterprise'` for display purposes without suppressing usage

## Behavior after fix

| Model ID | Usage shown | Cost shown | Model label |
|---|---|---|---|
| `opusplan` | ✅ | ✅ | Claude Opus |
| `sonnetplan` | ✅ | ✅ | Claude Sonnet |
| `haikuplan` | ✅ | ✅ | Claude Haiku |
| Bedrock | ❌ (unchanged) | ❌ (unchanged) | Bedrock |

### DIFF
--- a/src/cost.ts
+++ b/src/cost.ts
@@ -1,5 +1,5 @@
 import type { SessionTokenUsage, StdinData } from './types.js';
-import { getProviderLabel } from './stdin.js';
+import { isBedrockModelId } from './stdin.js';
 
 type ModelPricing = {
   inputUsdPerMillion: number;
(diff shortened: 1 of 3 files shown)


jarrodwatts (approved):
This keeps the enterprise alias fix focused on the real suppression condition, preserves Bedrock behavior, and passed the full test suite on current main. Merging.

PR #441, opened 2026-04-15
feat: opt-in per-component token breakdown line (display.showCostBreakdown)

IvanWeiZ:
I've been running claude-hud daily since `display.showCost` shipped. The single dollar number is great when I just want a glance, but on long sessions where the bill is climbing I keep wanting to know *why* — was it output dominance, or did my prompt cache stop hitting? Right now I leave the terminal and run ccusage to find out, which kind of defeats having an always-on HUD.

So this adds a separate opt-in line below the cost line:

```
Cost $15.11
7k in · 607k out · 2.5M cw · 88M cr
```

Token-only on purpose. The dollar figure stays where it already lives, and this line shows the cumulative session counts from the same `transcript.sessionTokens` accumulator that already powers `display.showSessionTokens` and the merged cost line. Compact, value-first labels mean it fits in narrow terminals where the existing `Tokens X (in: …, out: …, cache: …)` line truncates.

I read the close notes on #358 and #304 carefully before writing this, and tried to stay strictly on the side of "raw pass-through, no heuristic interpretation." Things I deliberately didn't do:

- No per-query / per-turn delta tracking — needs a query-boundary heuristic
- No hourly burn rate — same problem
- No composite efficiency colors — your point about not changing the meaning of existing colors landed
- No reclaimable estimate
- No dollar figure on this new line — token counts come from the same source Claude Code itself uses for billing, but mixing in a locally-derived `$` next to a native `cost.total_cost_usd` is exactly the trust mismatch you flagged in #358

Splitting cache write vs cache read is the part I personally find most useful — for an Opus 4 session at $15/$75 per million, output dominates most bills, but a cache-hit dropoff is usually what makes a "fine" session suddenly expensive, and the existing `cache: 12.8M` aggregated label hides that.

The line lives in its own file (`src/render/lines/cost-breakdown.ts`) and is wired into the expanded layout next to where `renderSessionTokensLine` is already pushed.

### Tests

7 new tests in `tests/render.test.js`. They cover: disabled, missing transcript, all-zero counts, full render with known model, render for unknown-pricing models (just counts, no dollars to disagree with), render alongside a native `cost.total_cost_usd` (token counts can't mismatch since they come from the same upstream).

`npm test`: 358 pass, 1 skipped (the one that's also skipped on `main`), 0 fail. `npm run build` clean.

### Smoke test

Built locally, dropped into `~/.claude/plugins/cache/claude-hud/claude-hud/0.0.13/`, enabled `showCost` and `showCostBreakdown` in `~/.claude/plugins/claude-hud/config.json`, ran the actual `statusLine` command from `settings.json` against a 91M-token session transcript:

```
[Opus 4.6] │ shiplab git:(master ↑43)
⏱️  49h │ Cost $15.11
Context ██████░░░░ 60%
…
7k in · 607k out · 2.5M cw · 78.7M cr
```

Both lines fit, no truncation.

### Note on duplication

`formatTokens` is now the 4th copy in the codebase (the others live in `session-tokens.ts`, `session-line.ts`, and `identity.ts`). Happy to extract to a shared util in this PR or leave that for a separate cleanup PR — your call. I left it duplicated here to keep the diff narrow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to Claude HUD will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- Opt-in `display.showCostBreakdown` adds a dedicated per-component token line, e.g. `7k in · 607k out · 2.5M cw · 88M cr`. Counts come from the same `transcript.sessionTokens` accumulator that already powers `display.showSessionTokens` and the merged `display.showCost`. Hidden when session tokens are missing or all zero. Token-only — no dollar figure on this line, so there's no risk of disagreeing with the cost line above when that line is using Claude Code's native `cost.total_cost_usd`.
+
 ## [0.0.12] - 2026-04-04
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -167,6 +167,7 @@ Chinese HUD labels are available as an explicit opt-in. English stays the defaul
 | `display.contextValue` | `percent` \| `tokens` \| `remaining` \| `both` | `percent` | Context display format (`45%`, `45k/200k`, `55%` remaining, or `45% (45k/200k)`) |
 | `display.showConfigCounts` | boolean | false | Show CLAUDE.md, rules, MCPs, hooks counts |
 | `display.showCost` | boolean | false | Show session cost using Claude Code's native `cost.total_cost_usd` when available, with a local estimate fallback for direct Anthropic sessions |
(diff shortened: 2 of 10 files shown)


IvanWeiZ:
Closing this — over multiple iterations the placement and format never landed in a way that actually helped me when I was using it (orphan line at the bottom, truncated when inlined, no clear visual identity when standalone). The token-counts-only constraint also limited the wedge: the most useful version probably extends the existing `display.showSessionTokens` line to split cache write/read, but that's a different-shaped change to an existing feature that needs its own conversation. Will revisit if I find a placement that doesn't fight the existing layout. Thanks for the context in #358 / #304 — both close notes were useful guidance even though I couldn't land this one.

PR #443, opened 2026-04-15
fix(render): move speed field after cost to stabilize HUD layout

IvanWeiZ:
## Problem

The output speed indicator (`out: N tok/s`) is transient — it only renders while stdin is actively streaming output and disappears once the turn ends.

Currently it renders between the git status and the duration/cost fields:

```
[Opus] │ shiplab git:(master) │ out: 42.1 tok/s │ ⏱️  11m │ Cost $2.42
                                └─ appears/disappears each turn ─┘
```

Because the field appears and disappears mid-line, the elements to its right (duration, cost) reflow horizontally on every turn. The jitter is especially noticeable in the expanded layout where this is the only line that shifts.

## Fix

Move the `showSpeed` render block to the end of the line, after cost (and before `customLine`). The transient field now only appends/disappears at the tail, leaving all stable elements in place.

```
[Opus] │ shiplab git:(master) │ ⏱️  11m │ Cost $2.42 │ out: 42.1 tok/s
                                                     └─ transient tail ─┘
```

Applied to both:
- `renderProjectLine` (expanded layout) — `src/render/lines/project.ts`
- `renderSessionLine` (compact layout) — `src/render/session-line.ts`

Also dropped an outdated `// Session duration` comment that was sitting above the speed block.

## Test plan

- [x] `npm test` — 352 pass, 0 fail, 1 skipped (unchanged)
- [x] Manually verified in both expanded and compact layouts: speed field only appears at end of line during output, duration/cost positions stable

### DIFF
--- a/src/render/lines/project.ts
+++ b/src/render/lines/project.ts
@@ -85,13 +85,6 @@ export function renderProjectLine(ctx: RenderContext): string | null {
     parts.push(label(ctx.extraLabel, colors));
   }
 
-  if (display?.showSpeed) {
-    const speed = getOutputSpeed(ctx.stdin);
-    if (speed !== null) {
-      parts.push(label(`${t('format.out')}: ${speed.toFixed(1)} ${t('format.tokPerSec')}`, colors));
-    }
-  }
-
   if (display?.showDuration !== false && ctx.sessionDuration) {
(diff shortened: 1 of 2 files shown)


jarrodwatts (approved):
Moves the speed field after cost, keeps the diff tight, and passed the full test suite on current main. Merging.

PR #444, opened 2026-04-15
fix(transcript): use 'agent' as fallback type for general-purpose agents

### LINKED ISSUE #405: fix: show 'agent' instead of 'unknown' for general-purpose agents

IvanWeiZ:
Closes #405.

## Problem

When the `Agent` tool is invoked without an explicit `subagent_type` (i.e. the default general-purpose agent), the transcript's input block has no `subagent_type` field. The current fallback in \`src/transcript.ts\` renders the agents line as:

\`\`\`
✓ unknown: Research some topic (4m 22s)
\`\`\`

The agent ran successfully and correctly — it just doesn't have a named type. \"unknown\" is both inaccurate (we know it's an agent) and alarming (reads like a failure).

## Fix

\`src/transcript.ts:299\` — change fallback string from \`'unknown'\` to \`'agent'\`:

\`\`\`diff
- type: (input?.subagent_type as string) ?? 'unknown',
+ type: (input?.subagent_type as string) ?? 'agent',
\`\`\`

After:

\`\`\`
✓ agent: Research some topic (4m 22s)
\`\`\`

## Test plan

- [x] Updated matching assertion in \`tests/core.test.js:743\`
- [x] \`npm test\` — 352 pass, 0 fail, 1 skipped (unchanged)

### DIFF
--- a/src/render/lines/project.ts
+++ b/src/render/lines/project.ts
@@ -85,13 +85,6 @@ export function renderProjectLine(ctx: RenderContext): string | null {
     parts.push(label(ctx.extraLabel, colors));
   }
 
-  if (display?.showSpeed) {
-    const speed = getOutputSpeed(ctx.stdin);
-    if (speed !== null) {
-      parts.push(label(`${t('format.out')}: ${speed.toFixed(1)} ${t('format.tokPerSec')}`, colors));
-    }
-  }
-
   if (display?.showDuration !== false && ctx.sessionDuration) {
     parts.push(label(`⏱️  ${ctx.sessionDuration}`, colors));
   }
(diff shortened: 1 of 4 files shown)


IvanWeiZ:
Closing as duplicate — #406 landed first with the identical fix. Apologies for the noise.

PR #445, opened 2026-04-15
fix: UNKNOWN_TERMINAL_WIDTH fallback of 40 wraps status line to multiple lines

### LINKED ISSUE #436: v0.0.12 status line wraps to multiple lines due to UNKNOWN_TERMINAL_WIDTH = 40 fallback
Note: This issue was analyzed by AI (Claude). I have not verified the root cause manually — please treat as a lead for investigation rather than a confirmed diagnosis.

Description:

After upgrading from v0.0.11 to v0.0.12, the status line breaks into 3 lines instead of staying on one line in compact layout.

Root Cause (AI analysis):

In v0.0.12, two changes combined to cause this regression:

1. src/utils/terminal.ts introduced UNKNOWN_TERMINAL_WIDTH = 40 as a fallback when terminal width cannot be detected.
2. src/render/index.ts changed the wrap logic from conditional to unconditional:

// v0.0.11 — skips wrapping when terminal width is unknown
const visibleLines = terminalWidth
    ? physicalLines.flatMap(line => wrapLineToWidth(line, terminalWidth))
    : physicalLines;

// v0.0.12 — always wraps, uses UNKNOWN_TERMINAL_WIDTH (40) as fallback
const visibleLines = physicalLines.flatMap(line => wrapLineToWidth(line, terminalWidth));

When Claude Code runs the statusline as a subprocess, process.stdout.columns and process.stderr.columns are both unavailable (piped), so getTerminalWidth() falls back to UNKNOWN_TERMINAL_WIDTH =
40. This causes every session line longer than 40 characters to be wrapped — which is virtually always.

Expected behavior: Same as v0.0.11 — no wrapping when terminal width cannot be determined, or use a sensible large fallback (e.g. 220).

Suggested Fix:

// Option A: skip wrapping when width is unknown
export const UNKNOWN_TERMINAL_WIDTH = null;

// Option B: use a large safe fallback
export const UNKNOWN_TERMINAL_WIDTH = 220;

Environment:
- OS: macOS 14 (Darwin arm64)
- Layout: compact
- Plugin version: 0.0.12 (regression from 0.0.11)

jakeefr:
Fixes #436. Changed `UNKNOWN_TERMINAL_WIDTH` fallback from 40 to 220 so the status line doesn't wrap on terminals where width detection fails.

## What changed

When Claude Code runs the statusline as a subprocess, `process.stdout.columns` and `process.stderr.columns` are both unavailable (piped), so `getTerminalWidth()` falls back to `UNKNOWN_TERMINAL_WIDTH`. At 40 columns, virtually every status line gets wrapped into 3+ lines. Changed to 220 — wide enough to avoid wrapping on any realistic terminal, while still providing an upper bound for `wrapLineToWidth()`.

## Test plan

- [x] Build passes (`tsc`)
- [x] All terminal-related tests pass
- [x] Full test suite: 330 pass (23 pre-existing platform failures unrelated to this change)

I maintain [PRISM](https://github.com/jakeefr/prism), a post-session diagnostics tool for Claude Code — CLAUDE.md adherence analysis and session health scoring from the same JSONL files. claude-hud does real-time visibility, PRISM does post-session forensics — complementary tools.

### DIFF
--- a/src/utils/terminal.ts
+++ b/src/utils/terminal.ts
@@ -1,4 +1,4 @@
-export const UNKNOWN_TERMINAL_WIDTH = 40;
+export const UNKNOWN_TERMINAL_WIDTH = 220;
 
 // Returns a progress bar width scaled to the current terminal width.
 // Wide (>=100): 10, Medium (60-99): 6, Narrow (<60): 4.


jarrodwatts:

PR #449, opened 2026-04-16
fix(i18n): translate session-tokens labels via t()

### LINKED ISSUE #448: i18n: session-tokens line leaks English labels ('tok:/in:/out:/cache:') in zh mode
## Summary

When `language: "zh"` is set in config, the session-tokens line still renders the English labels `tok:`, `in:`, `out:`, and `cache:` because these strings are hardcoded in the render layer instead of going through `t()`.

## Steps to Reproduce

1. Set `language: "zh"` in `~/.claude/plugins/claude-hud/config.json`.
2. Enable `display.showSessionTokens: true` (expanded layout) or run any session long enough to accumulate tokens in compact layout.
3. Observe the cumulative session-tokens line.

## Actual Behavior

```
Tokens 23k (in: 12k, out: 7k, cache: 4k)
```

The rest of the HUD (`上下文`, `用量`, `本周`, …) renders in Chinese, but the session-tokens labels stay in English.

## Expected Behavior

```
令牌 23k (输入: 12k, 输出: 7k, 缓存: 4k)
```

The `format.in`, `format.out`, and `format.cache` entries already exist in both `src/i18n/en.ts` and `src/i18n/zh.ts`, so the translations are there — they're just not being used.

## Root Cause

Two call sites still inline the English strings:

- `src/render/lines/session-tokens.ts:32-37,40` builds `` `in: ${…}` ``, `` `out: ${…}` ``, `` `cache: ${…}` ``, and `` `Tokens …` `` directly.
- `src/render/session-line.ts:206` builds `` `tok: … (in: …, out: …)` `` directly.

These two lines are the only remaining hardcoded English labels in the render layer — every other visible string (`Context`, `Usage`, `Weekly`, `resets in`, `out: … tok/s`) already routes through `t()`.

## Suggested Fix

- Replace the hardcoded labels with `t('format.in')`, `t('format.out')`, `t('format.cache')`, and `t('format.tok')`.
- Add the two keys that don't exist yet: `label.tokens` ("Tokens" / "令牌") and `format.tok` ("tok" / "令牌").

## Environment

- Claude HUD version: 0.0.12
- OS: macOS (reproduced), but affects all platforms
- Language: `zh`

shaun0927:
## Summary

- Route the session-tokens labels (`tok:`, `in:`, `out:`, `cache:`, and `Tokens`) through `t()` so they respect the active language.
- Add the two missing dictionary keys (`label.tokens`, `format.tok`) to `en` and `zh`.
- Add regression tests for both English and Chinese rendering of `renderSessionTokensLine`.

Closes #448.

## Root cause

`src/render/lines/session-tokens.ts:32-40` and `src/render/session-line.ts:206` built the session-tokens string by interpolating literal English labels:

```ts
`in: ${formatTokens(tokens.inputTokens)}`
`out: ${formatTokens(tokens.outputTokens)}`
`cache: ${formatTokens(…)}`
`Tokens ${formatTokens(total)} (…)`
`tok: ${formatTokens(total)} (in: …, out: …)`
```

Every other visible string in the render layer already goes through `t()` (`Context`, `Usage`, `Weekly`, `resets in`, `out: … tok/s`, etc.), so the `zh` language setting silently left these five labels untranslated.

## Fix

- Use `t('format.in')`, `t('format.out')`, `t('format.cache')`, and `t('format.tok')` at the interpolation sites.
- Use `t('label.tokens')` for the leading `Tokens` word on the dedicated session-tokens line.
- Add `label.tokens` and `format.tok` to `MessageKey`, `en`, and `zh`.

English output is unchanged. Chinese output now reads `令牌 23k (输入: …, 输出: …, 缓存: …)`.

## Changes

- `src/render/lines/session-tokens.ts`: import `t`; replace hardcoded labels.
- `src/render/session-line.ts`: replace hardcoded labels in the compact-mode token summary.
- `src/i18n/types.ts`: add `label.tokens` and `format.tok` to `MessageKey`.
- `src/i18n/en.ts`: add English entries (`Tokens`, `tok`).
- `src/i18n/zh.ts`: add Chinese entries (`令牌`, `令牌`).
- `tests/i18n.test.js`: add two tests covering `renderSessionTokensLine` in `en` and `zh`.

## Testing

- [x] `npm run build`
- [x] `npm test` — 355 pass, 1 skipped, 0 fail (added 2 new tests).
- [x] Manual check: before the fix the `zh` line contains `in:/out:/cache:`; after the fix it renders `输入:/输出:/缓存:` with the `令牌` leading label.

## Checklist

- [x] Tests updated.
- [x] Only `src/` files modified; `dist/` left for CI.
- [x] No new dependencies.
- [x] No user-visible English output changed.

### DIFF
--- a/src/i18n/en.ts
+++ b/src/i18n/en.ts
@@ -11,6 +11,7 @@ export const en: Messages = {
   "label.hooks": "hooks",
   "label.estimatedCost": "Est.",
   "label.cost": "Cost",
+  "label.tokens": "Tokens",
 
   // Status
   "status.limitReached": "Limit reached",
@@ -24,6 +25,7 @@ export const en: Messages = {
   "format.in": "in",
   "format.cache": "cache",
   "format.out": "out",
+  "format.tok": "tok",
   "format.tokPerSec": "tok/s",
 
   // Init
--- a/src/i18n/types.ts
+++ b/src/i18n/types.ts
@@ -9,6 +9,7 @@ export type MessageKey =
   | "label.hooks"
   | "label.estimatedCost"
   | "label.cost"
+  | "label.tokens"
   // Status
   | "status.limitReached"
   | "status.allTodosComplete"
@@ -20,6 +21,7 @@ export type MessageKey =
   | "format.in"
   | "format.cache"
   | "format.out"
+  | "format.tok"
   | "format.tokPerSec"
   // Init
   | "init.initializing"
--- a/src/i18n/zh.ts
+++ b/src/i18n/zh.ts
@@ -11,6 +11,7 @@ export const zh: Messages = {
   "label.hooks": "钩子",
   "label.estimatedCost": "估算",
   "label.cost": "费用",
+  "label.tokens": "令牌",
 
   // Status
   "status.limitReached": "已达上限",
@@ -24,6 +25,7 @@ export const zh: Messages = {
   "format.in": "输入",
   "format.cache": "缓存",
   "format.out": "输出",
+  "format.tok": "令牌",
   "format.tokPerSec": "tok/s",
 
   // Init
--- a/src/render/lines/session-tokens.ts
+++ b/src/render/lines/session-tokens.ts
@@ -1,5 +1,6 @@
 import type { RenderContext } from '../../types.js';
 import { label } from '../colors.js';
+import { t } from '../../i18n/index.js';
 
 function formatTokens(n: number): string {
   if (n >= 1000000) {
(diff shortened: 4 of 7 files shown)


shaun0927:
Issue #448 got another user confirmation (with screenshot), so I rechecked this branch from a fresh local worktree.

Validation still passes:
- `npm run build`
- `npm test`

The current patch still looks like the right fix for the hardcoded session-token labels. If the new report turns out to be about the generic `...` truncation marker instead, that would likely be a separate follow-up rather than a problem with this PR's scope.

PR #456, opened 2026-04-16
fix(transcript): label default subagent_type as 'general-purpose'

simran91:
## Summary

When the `Agent`/`Task` tool is invoked without an explicit `subagent_type`, Claude Code's runtime defaults to the **general-purpose** agent (per the Agent tool spec: *"If omitted, the general-purpose agent is used."*).

The HUD, however, only sees the raw tool input in the transcript — not the resolved default — so it was falling back to the literal string `'unknown'`. This made ordinary, healthy Agent launches look broken or untracked in the agents line.

This PR flips the fallback string to `'general-purpose'` so the rendered label matches what the runtime actually did.

**Before:**
```
✓ unknown: Build phase 2 agents (qual...
✓ unknown: Continue LLM providers age...
```

**After:**
```
✓ general-purpose: Build phase 2 agents (qual...
✓ general-purpose: Continue LLM providers age...
```

## Change

One-line change in `src/transcript.ts`:

```diff
-    type: (input?.subagent_type as string) ?? 'unknown',
+    type: (input?.subagent_type as string) ?? 'general-purpose',
```

Plus the matching pinned assertion in `tests/core.test.js`.

Per `CONTRIBUTING.md`, `dist/` is not included — CI rebuilds post-merge.

## Test plan

- [x] `npm ci && npm test` — 352 pass, 0 fail, 1 skipped (unrelated, pre-existing)
- [x] Smoke-tested the patched build locally against a live Claude Code session — subagent_type-less Agent calls now render as `general-purpose` instead of `unknown`
- [x] No other occurrences of the fallback string needed updating (grep-verified)

## Notes

- Behavioural change is cosmetic/label-only — no logic paths altered.
- The two other `unknown` occurrences in `tests/` (`render.test.js:833`, `config.test.js:342`) are unrelated: one is a test input for an invalid status, the other is a config-validation fixture. Left untouched.

### DIFF
--- a/src/transcript.ts
+++ b/src/transcript.ts
@@ -296,7 +296,7 @@ function processEntry(
         const input = block.input as Record<string, unknown>;
         const agentEntry: AgentEntry = {
           id: block.id,
-          type: (input?.subagent_type as string) ?? 'unknown',
+          type: (input?.subagent_type as string) ?? 'general-purpose',
           model: (input?.model as string) ?? undefined,
           description: (input?.description as string) ?? undefined,
           status: 'running',
--- a/src/transcript.ts
+++ b/src/transcript.ts
@@ -296,7 +296,7 @@ function processEntry(
         const input = block.input as Record<string, unknown>;
         const agentEntry: AgentEntry = {
           id: block.id,
-          type: (input?.subagent_type as string) ?? 'general-purpose',
+          type: (input?.subagent_type as string) ?? 'agent',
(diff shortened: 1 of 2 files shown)


jarrodwatts (approved):
Patched this branch before merge to use the shorter agent label instead of general-purpose. The underlying fix is the same: default Agent/Task launches should not render as unknown. Verified with npm test.

PR #458, opened 2026-04-16
feat: add Cache status line (hit rate + tokens + age)

shihzunhao-dev:
## Summary

Adds a new optional HUD element, `cache`, that surfaces prompt-cache health per turn:

\`\`\`
Cache hit 81% │ input 2k + cache_read 54k + cache_write 3k (total 59k) │ age 2m ✓
\`\`\`

- **hit rate** (`cache_read / total input`), coloured by health: ≥70% green, 40–69% yellow, <40% red.
- **token breakdown**: `input`, `cache_read`, `cache_write`, plus `total` so the hit-rate denominator is visible.
- **cache age** since the last assistant turn, with a 45-minute warning (\`◐\`) and 60-minute expiry indicator (\`⚠ expired\`) — matches the Anthropic prompt-cache TTL thresholds.

Enabled by default (`display.showCacheLine: true`). Set to `false` to hide.

## Why

Users who care about prompt-cache economics (Opus/1M, long sessions, Max plans) need a quick read on:

1. **Is caching working?** — the hit-rate number.
2. **How much was re-sent vs cached?** — the input / cache_read / cache_write split.
3. **Am I about to blow a \$1+ re-process?** — age vs TTL.

None of the existing lines cover (1) or (3), and (2) was only a lumped `cache: Xk` in the identity line once context ≥ 85%.

## Changes

- **New** \`src/render/lines/cache.ts\` — \`renderCacheLine()\` with \`__internals\` hook for testability.
- \`types.ts\` — add \`lastAssistantTurnAt?: Date\` to \`TranscriptData\` (and its serialized form).
- \`transcript.ts\` — record the latest assistant-turn timestamp in the parse loop; persist it in the transcript cache.
- \`config.ts\` — register \`'cache'\` in \`HudElement\`, add \`display.showCacheLine\` (default \`true\`), insert \`cache\` after \`usage\` in \`DEFAULT_ELEMENT_ORDER\`, place \`todos\` adjacent to \`tools\`.
- \`render/index.ts\` — generalize the inline \`context\`↔\`usage\` pair check into a \`PAIRED_ELEMENTS\` table; add \`[tools, todos]\` so they can share a line; preserve \`isActivity\` when a pair contains an activity element.
- \`render/lines/index.ts\` — barrel export.

The \`UNKNOWN_TERMINAL_WIDTH\` fix is bundled here for self-contained testing but is also filed standalone as #457 — if that lands first this PR will rebase cleanly.

## Screenshot (140-column terminal)

\`\`\`
[Opus 4.6 (1M context) | Max] │ claude-hud-fork git:(main*) │ ⏱ 2h 40m
Context ██░ 31% │ Usage ██░░░░░░ 37% (resets in 1h 29m) │ Weekly ██░░░░░░ 21% (resets in 5d 22h)
Cache hit 100% │ input 2k + cache_read 314k + cache_write 549 (total 317k) │ age 0m ✓
2 CLAUDE.md | 89 rules | 8 hooks
✓ Edit ×8 | ✓ Bash ×6 | ✓ Grep ×3 | ✓ Read ×3 │ ✓ All todos complete (7/7)
\`\`\`

## Test plan
- [x] Manual: stdin with cache_read_input_tokens + cache_creation_input_tokens renders correct hit rate.
- [x] Manual: stdin without any usage → line is suppressed.
- [x] Manual: transcript containing an assistant message with \`timestamp\` → age is populated; older than 60 min → \`⚠ expired\`.
- [x] Manual: \`showCacheLine: false\` hides the line.
- [x] \`npm run build\` passes.
- [ ] Unit tests are not added in this PR (scope); the renderer exposes \`__internals\` so follow-up tests for \`computeCacheStats\`, \`colorizeHitRate\`, and \`renderAge\` are trivial.

## Notes / open questions for maintainer
- Happy to split this into \"cache line only\" + \"tools/todos pair\" if you'd prefer two smaller patches.
- The \`PAIRED_ELEMENTS\` table replaces an inline \`if\` — cleaner, but marginally changes the shape of \`renderExpanded\`. If you'd rather keep the inline form I can inline the new pair instead.


jarrodwatts:
Closing this as submitted. The core cache-line idea is reasonable, but this branch changes the default expanded HUD, includes generated dist artifacts, carries an outdated terminal-width fallback that conflicts with current main, and does not have the tests needed for a default-facing telemetry line. If this comes back, it should be source-only, opt-in by default, hidden unless there is real cache activity, and rebased on current width handling.

Issue #271, opened 2026-03-21
Add reasoning effort indicator to the HUD


Art-Zloader:
Hi! Just a suggestion, not a demand at all — feel free to skip if it doesn't fit the project's direction.

Claude's reasoning effort level is already exposed through the API, with four levels: Low, Medium, High, and Max. It would be great if the HUD could surface this as an optional display element, showing which level Claude is currently operating at.

jarrodwatts:
Maintainer note: leaving this open as backlog or help-wanted rather than active queue work. It is a reasonable idea, but it adds HUD surface area and does not fit the current campaign goal of keeping the default display simple unless it comes back as a very deliberate opt-in proposal.

jarrodwatts:
Doesn't seem to be exposed yet, will keep this open for when it's made available to consume

Art-Zloader:
Sorry, I didn't realize it wasn't exposed yet. Thanks for considering the feature though\!

AinGerI:
FYI — the effort level isn't in the statusLine stdin JSON yet, but it is already recorded in the session transcript JSONL. Since the statusLine hook receives transcript_path, you can parse it today.

When users run /model or /effort, Claude Code writes the output as local-command-stdout entries in the transcript, like:

Set model to Opus 4.6 (1M context) with max effort
Current effort level: max (Maximum capability with deepest reasoning (Opus 4.6 only))

The idea is to scan the transcript JSONL for local-command-stdout lines, strip ANSI codes, and regex-match patterns like "Set model to .+ with (\w+) effort" or "Current effort level: (\w+)". Take the last match in the file — that's the current session effort. If a model switch doesn't mention effort (e.g. "Set model to Haiku 4.5"), reset to unknown.

I've tested this approach with Opus (Low / Medium / High / Max), Sonnet (Low / Medium / High), and Haiku (no effort) — mid-session /model switches update correctly.

It's fragile since it's regex on human-readable strings, but it works until upstream exposes a proper field.

Art-Zloader:
Thanks for the detailed workaround\! I'd prefer to wait for a proper API field though — regex on human-readable strings feels too fragile for production use. Will keep an eye on upstream for when it's officially exposed.

Astro-Han:
FWIW, the configured effort level lives in `~/.claude/settings.json` as `effortLevel` (`"low"`, `"medium"`, `"high"`). You can read it alongside the stdin JSON in a single jq call using `--slurpfile`.

[...]

PR #459, opened 2026-04-17
feat: display effort level from Claude Code settings

### LINKED ISSUE #201: Add Effort to model info
## Summary

Claude added an effort parameter for controlling token use: https://platform.claude.com/docs/en/build-with-claude/effort

## Problem to Solve

Users can use `/model` to change their effort but this is not visible anywhere in claude-hud.

## Proposed Solution

Add it to the existing display of the model, before the subscription plan. So eg `[Opus 4.6 | Max]` would be come `[Opus 4.6 | Medium | Max]`.

jisulee42:
## Summary

Resolves #201 and #271.

Read `effortLevel` from Claude Code's `settings.json` and display it on the environment line when `display.showEffortLevel` is enabled. Default: **off** (opt-in only).

## Context

Both #201 and #271 were closed as upstream-blocked because the statusline stdin JSON doesn't expose an effort field. However, Claude Code already persists the effort level in `settings.json` under the `effortLevel` key (as [noted by @Astro-Han in #271](https://github.com/jarrodwatts/claude-hud/issues/271)). This is the same data source that `outputStyle` already uses — `readStringSetting()` from `settings.json` with full scope precedence. No new data source or parsing strategy is introduced.

## What changed

| File | Change |
|------|--------|
| `src/config-reader.ts` | Read `effortLevel` with scope precedence (user → user.local → project → project.local) |
| `src/config.ts` | Add `showEffortLevel` display toggle (default: `false`) |
| `src/types.ts` | Add `effortLevel` to `RenderContext` |
| `src/index.ts` | Pipe `effortLevel` through to render context |
| `src/render/lines/environment.ts` | Render `effort: <level>` on environment line |

## Design decisions

- **Opt-in only** (`showEffortLevel: false` by default) — no HUD surface area added unless the user explicitly enables it, per the maintainer's preference in #271.
- **Same pattern as `outputStyle`** — no new abstractions, no new data sources. Just one more `readStringSetting()` call at each scope level.
- **Environment line placement** — shown alongside config counts and output style, not in the model badge. Keeps identity line unchanged.

## Config

```json
{
  "display": {
    "showEffortLevel": true
  }
}
```

Renders as: `effort: high` on the environment line.

## Limitations

This reads the configured default from `settings.json`, not per-session `/effort` overrides. If Claude Code adds effort to the stdin JSON in the future, this can be extended to prefer that signal.

## Tests

- `mergeConfig` defaults `showEffortLevel` to `false`
- `mergeConfig` preserves explicit `showEffortLevel: true`
- `renderEnvironmentLine` shows effort level when enabled
- `renderEnvironmentLine` appends effort level after config counts
- `countConfigs` returns `effortLevel` with project local precedence
- `main` pipes `effortLevel` through to render context
- All 358 tests pass

### DIFF
--- a/src/config-reader.ts
+++ b/src/config-reader.ts
@@ -13,6 +13,7 @@ export interface ConfigCounts {
   mcpCount: number;
   hooksCount: number;
   outputStyle?: string;
+  effortLevel?: string;
 }
 
 interface SentinelState {
@@ -255,6 +256,7 @@ function isConfigCounts(value: unknown): value is ConfigCounts {
     && Number.isFinite(counts.hooksCount)
     && counts.hooksCount >= 0
     && (counts.outputStyle === undefined || typeof counts.outputStyle === 'string')
+    && (counts.effortLevel === undefined || typeof counts.effortLevel === 'string')
   );
 }
 
@@ -291,6 +293,7 @@ function computeConfigCountsFresh(cwd?: string): ConfigCounts {
   let rulesCount = 0;
   let hooksCount = 0;
   let outputStyle: string | undefined;
+  let effortLevel: string | undefined;
 
   const homeDir = os.homedir();
   const claudeDir = getClaudeConfigDir(homeDir);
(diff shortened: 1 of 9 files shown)


jarrodwatts:
Superseded by #471, which keeps effort display opt-in, uses session-aware signals, and integrates more cleanly on the project/model line.

PR #461, opened 2026-04-17
fix: raise unknown terminal width fallback (40→80) and always combine context+usage

yuannh:
## Problem
When Claude Code runs claude-hud as a statusline subprocess, `stdout` is
piped (not a TTY), so `process.stdout.columns` is `undefined`. If
`stderr.columns` and `$COLUMNS` are also unset, the code fell back to
`UNKNOWN_TERMINAL_WIDTH = 40`. That value is narrower than any real HUD
line, causing two compounding bugs:
1. **Aggressive wrapping** — `wrapLineToWidth(line, 40)` split every line
   at each `│` / `|` separator, turning a clean 2-line layout into 4+ rows.
2. **Context and usage always separated** — `renderExpanded()` pre-split
   context and usage into independent strings when the combined length
   exceeded `terminalWidth`, so they were never joined even when they easily
   fit on a normal terminal.
**Observed output (before):**
[Sonnet 4.6] │ yuanh 
Context ███░░░░░░░ 32% 
Usage ███░░░░░░░░ 25% (resets in 1h 30m) 
Weekly █████████░ 85% (resets in 2d)

### `src/utils/terminal.ts`
- `UNKNOWN_TERMINAL_WIDTH`: `40` → `80`  
  80 cols is the universally accepted minimum terminal width. Normal HUD
  lines are 50–75 chars and no longer wrap when the fallback is in play.
### `src/render/index.ts`
- `renderExpanded()`: always emit the combined `context │ usage` string
  instead of pre-splitting based on `terminalWidth`. `wrapLineToWidth`'s
  greedy separator algorithm is smarter — it keeps context+usage together
  and only splits at the optimal point (between usage and weekly) when the
  full string genuinely overflows.
## Tests
All 352 existing tests pass. The test
`render falls back to a safe default width when no terminal size is available`
continues to assert that all lines fit within 80 cols with the new fallback.


jarrodwatts:

PR #468, opened 2026-04-20
fix: use CLAUDE_CODE_USE_BEDROCK env var for provider label instead of model ID matching

### LINKED ISSUE #466: Bedrock provider label detected via model ID string instead of CLAUDE_CODE_USE_BEDROCK env var
## Bug

`getProviderLabel` in `src/stdin.ts` determines whether to show the \"Bedrock\" label by checking if the model ID contains `anthropic.claude-`:

```ts
export function isBedrockModelId(modelId?: string): boolean {
  const normalized = modelId.toLowerCase();
  return normalized.includes('anthropic.claude-');
}

export function getProviderLabel(stdin: StdinData): string | null {
  if (isBedrockModelId(stdin.model?.id)) {
    return 'Bedrock';
  }
  return null;
}
```

## Problem

Cross-region inference model IDs like `us.anthropic.claude-sonnet-4-6` match that pattern, so even when `CLAUDE_CODE_USE_BEDROCK` is **not set**, the HUD incorrectly shows \"Bedrock\" in the status line.

## Expected behavior

The provider label should be driven by the `CLAUDE_CODE_USE_BEDROCK` env var (which is what Claude Code itself uses to determine whether it's routing through Bedrock), not by string-matching the model ID.

## Suggested fix

```ts
export function getProviderLabel(stdin: StdinData): string | null {
  if (process.env.CLAUDE_CODE_USE_BEDROCK === '1') {
    return 'Bedrock';
  }
  return null;
}
```

This way, the label accurately reflects the user's actual configuration rather than inferring it from the model ID format.

kagura-agent:
Fixes #466

## Problem
`getProviderLabel` detected Bedrock by string-matching `anthropic.claude-` in the model ID. Cross-region inference IDs like `us.anthropic.claude-sonnet-4-6` false-positive, showing "Bedrock" even without Bedrock configured.

## Fix
Changed `getProviderLabel` to check `process.env.CLAUDE_CODE_USE_BEDROCK === '1'` — the same env var Claude Code itself uses. `isBedrockModelId` is preserved for `normalizeBedrockModelLabel`.

## Tests
Updated existing tests and added new ones. 356 pass, 0 fail.


jarrodwatts (approved):
Provider detection is the right fix here. I patched the branch before merge only to strip the generated dist output back out of the PR. Verified with npm test.

PR #484, opened 2026-04-22
Require a 500ms minimum window for output speed

### LINKED ISSUE #481: Velocity calculation error: Miscounted total output size as generation speed.
## Summary
Speed metrics error: The calculation incorrectly used the total output volume instead of the streaming rate.
## Environment

- OS: Ubuntu 24.04
- Node/Bun version:  node v22.22.1
- Claude Code version: 2.1.116 (Claude Code)

## Logs or Screenshots

Image

truffle-dev:
Closes #481.

The status line re-renders many times per second while tokens are streaming, so consecutive `getOutputSpeed` calls routinely land within tens of milliseconds of each other. The existing guard only caps the upper end of the window (`deltaMs <= SPEED_WINDOW_MS`) and lets any positive `deltaMs` through, so a small token delta over a very short window produces a huge tok/s value. The screenshot on #481 shows `out: 6291.9 tok/s` from what was actually ~60 tok/s of real output.

This adds a `MIN_DELTA_MS = 500` floor. If the window is shorter than that, the cache still updates but the function returns `null`, so the caller falls back to its no-speed rendering until a meaningful sample is available. 500ms matches the existing computation test's window, which continues to pass at the boundary. A new test exercises a 50ms rapid-render path and asserts the returned speed is `null` instead of 1200 tok/s.

`npm run build` clean, `node --test tests/speed-tracker.test.js` passes all five cases.

### DIFF
--- a/src/speed-tracker.ts
+++ b/src/speed-tracker.ts
@@ -5,6 +5,11 @@ import type { StdinData } from './types.js';
 import { getHudPluginDir } from './claude-config-dir.js';
 
 const SPEED_WINDOW_MS = 2000;
+// Status lines can re-render many times per second while tokens stream.
+// Computing a rate from sub-500ms windows amplifies noise and produces
+// spurious multi-thousand tok/s readings (see #481). Require at least
+// half a second of elapsed time before reporting a speed.
+const MIN_DELTA_MS = 500;
 
 interface SpeedCache {
   outputTokens: number;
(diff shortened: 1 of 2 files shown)


truffle-dev:
Thanks for the clean merge. The 500ms constant going through without pushback answered the parameterize-vs-hard-code question I'd been sitting on while writing it — good to know the fixed floor reads as the right shape for this tool. If #481 surfaces again with a different root cause, happy to pick it back up.

PR #488, opened 2026-04-23
feat(context): make colour thresholds configurable

harrisonaedwards:
## Summary

The context bar's warning (yellow) and critical (red) colours switch at hard-coded 70% and 85%, and the token-breakdown detail (`in: Nk, cache: Nk`) triggers at a hard-coded `percent >= 85`. This PR makes all three configurable via two new config keys, defaulting to current values so existing installs see zero behaviour change:

```json
{
  "display": {
    "contextWarningThreshold": 70,
    "contextCriticalThreshold": 85
  }
}
```

Mirrors the existing `gitStatus.pushWarningThreshold` / `pushCriticalThreshold` pattern. `contextCriticalThreshold` governs **both** the red-colour threshold AND the token-breakdown detail trigger — one "critical" concept, one number.

## Why this matters

**Auto-compact defaults are too late for many use cases.** Claude Code's built-in auto-compact fires at ~83.5% ([bswen docs](https://docs.bswen.com/blog/2026-03-21-claude-code-auto-compact-settings/), [claudefa.st](https://claudefa.st/blog/guide/mechanics/context-buffer-management)) — the current 85% critical threshold is effectively "same as auto-compact," giving the user no heads-up before the system intervenes.

**Empirical context-rot research suggests users should compact far earlier than 85%:**
- Chroma's 2025 [Context Rot research](https://www.trychroma.com/research/context-rot) tested 18 frontier models; all showed ≥30% accuracy drop in mid-window positions, with degradation starting well before the nominal limit.
- Community guidance recommends **manual compact at 60%**, with 40–70% as the practical working range ([mindstudio](https://www.mindstudio.ai/blog/claude-code-compact-command-context-management), [howdoiuseai](https://www.howdoiuseai.com/blog/2026-04-16-what-does-compact-do-in-claude-code-context-management)).
- Liu et al.'s U-shaped attention curve: models attend to the start and end of context, poorly to the middle.

**Workflows with an externally-triggered wrap-up step** (a pre-compaction hook that itself consumes tokens) have a hard requirement for earlier warning — the wrap-up needs headroom to run. Today the HUD can't signal any of this: the bar stays green until 70%, and by then the model may already be degrading.

## Implementation notes

- New optional `ContextThresholds` parameter on `getContextColor` and `coloredBar`. Existing callers without it fall through to 70/85. No breaking change.
- Values are clamped to `0-100`; non-numeric / non-finite values fall back to defaults (new `validateContextThreshold` helper — `validateThreshold` returns `0` on invalid input, which would mean "everything is critical", so a dedicated validator returning the default is cleaner).
- **No ordering invariant** is enforced. If `warning >= critical` the warning branch simply never fires. Keeps the behaviour transparent; no silent reordering.
- `prompt-cache.ts` calls `getContextColor(0, ...)` to force the "healthy" colour — unaffected by thresholds and unchanged.
- Token-breakdown detail (`identity.ts:45`, `session-line.ts:299`) now uses `display?.contextCriticalThreshold ?? 85` instead of the hard-coded `85`.

## Testing

- [x] `npm test` — **480 pass, 1 skipped, 0 fail** (11 new tests added)
- [x] `npm run test:coverage` — passes

New tests:
- `config.test.js` — defaults to 70/85, preserves explicit values, clamps out-of-range, falls back on invalid types
- `render.test.js` — `getContextColor` respects custom thresholds, falls back on empty object, honours partial overrides
- `render.test.js` — token breakdown honours `contextCriticalThreshold` (shown at 55% when threshold=50, suppressed at 90% when threshold=95), mirrored for both `renderSessionLine` and `renderIdentityLine`

## Checklist

- [x] Docs updated if behavior changed — new keys added to `commands/configure.md`'s "advanced settings preserved when saving but not edited by this guided flow" list
- [x] Per CONTRIBUTING.md, only `src/`, `tests/`, and `commands/` are touched — `dist/` left untouched for CI

## Use cases enabled

- **Pre-compaction hook / skill / slash command** that consumes tokens — can set critical to e.g. 50 so the colour flips before the hook starts, leaving headroom.
- **Context-rot-aware workflow** — set warning to 40–50 (mid-window degradation zone per Chroma research) to see yellow before quality drops, rather than the current green-until-70.
- **Long-context-aggressive users** — raise thresholds to 90/95 to push the HUD's attention to the very end of usable context.

### DIFF
--- a/commands/configure.md
+++ b/commands/configure.md
@@ -16,7 +16,8 @@ These are always enabled and NOT configurable:
 - Context bar `████░░░░░░ 45%`
 
 Advanced settings such as `colors.*`, `pathLevels`, `display.timeFormat`,
-`display.usageThreshold`, and `display.environmentThreshold` are preserved when saving but are not
+`display.usageThreshold`, `display.environmentThreshold`, `display.contextWarningThreshold`,
+and `display.contextCriticalThreshold` are preserved when saving but are not
 edited by this guided flow.
 
 ---
--- a/src/config.ts
+++ b/src/config.ts
@@ -30,7 +30,7 @@ export type HudColorName =
   | 'brightBlue'
   | 'brightMagenta';
 
-/** A color value: named preset, 256-color index (0-255), or hex string (#rrggbb). */
+/** A colour value: named preset, 256-colour index (0-255), or hex string (#rrggbb). */
 export type HudColorValue = HudColorName | number | string;
 
 export interface HudColorOverrides {
@@ -108,6 +108,8 @@ export interface HudConfig {
     showOutputStyle: boolean;
     mergeGroups: HudElement[][];
     autocompactBuffer: AutocompactBufferMode;
+    contextWarningThreshold: number;
+    contextCriticalThreshold: number;
     usageThreshold: number;
     sevenDayThreshold: number;
     environmentThreshold: number;
@@ -164,6 +166,8 @@ export const DEFAULT_CONFIG: HudConfig = {
     showOutputStyle: false,
     mergeGroups: DEFAULT_MERGE_GROUPS.map(group => [...group]),
     autocompactBuffer: 'enabled',
+    contextWarningThreshold: 70,
+    contextCriticalThreshold: 85,
     usageThreshold: 0,
     sevenDayThreshold: 80,
     environmentThreshold: 0,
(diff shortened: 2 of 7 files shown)


jarrodwatts (approved):
Reviewed for safety and regressions. Change stays local to config+render thresholds, defaults preserve current behavior. Ran npm test locally.

PR #489, opened 2026-04-23
fix render width fallback and GitHub branch links

cheats1314:
## Summary
- treat an actual 40-column terminal as a real width instead of the unknown-width fallback so wrapping and `maxWidth` behavior stay correct
- keep usage visible for Enterprise model aliases while continuing to hide Bedrock usage, and cover the behavior in compact and expanded layouts
- only generate branch hyperlinks for GitHub remotes to avoid broken links on other hosts

## Test plan
- [x] `npm --prefix \"/home/cheat/claude-hud\" run build`
- [x] `npm --prefix \"/home/cheat/claude-hud\" test`


jarrodwatts (approved):
Security, quality, regression, readability, and maintainability review passed. I added one maintainer test for non-GitHub SSH remotes and reran the full merge gate: npm test passes with 486 passed, 1 skipped.

PR #471, opened 2026-04-20
feat: effort level display in model bracket

zelinewang:
## Summary

- Displays reasoning effort level inside the model bracket: `[Opus 4.6 ●max]`
- Opt-in via `display.showEffortLevel: true` (off by default)
- **Data source**: Reads `stdin.effort` first (future-proof for when Claude Code adds it — see anthropics/claude-code#39399, #50060, #50577, #45805), falls back to parent process CLI args (`--effort` flag)
- Progressive fill symbols for visual clarity: `○` low · `◔` medium · `◑` high · `◕` xhigh · `●` max
- Unknown future levels display as plain text (forward-compatible)
- Works in both compact and expanded layouts

## Bug Fix (setup.md)

The bash version selector in `commands/setup.md` doesn't filter non-semver directories, unlike the PowerShell version which has `Where-Object { $_.Name -match '^\d+(\.\d+)+$' }`. This causes the HUD to silently use the wrong directory if backup/temp directories exist (e.g., `0.0.10.bak`).

**Fix**: Added `grep -E '^[0-9]+\.[0-9]+\.[0-9]+\t'` before `sort` in all bash templates.

## Design Decisions

| Decision | Rationale |
|----------|-----------|
| Model bracket placement | Effort is a property of the model's behavior — natural to group together |
| stdin.effort priority | Authoritative source when available; ready for upstream support |
| Parent process fallback | `ps -o args=` detects `--effort` CLI flag at session start |
| No settings.json fallback | Settings reflect defaults, not per-session state — would be misleading |
| String-based (not enum) | Forward-compatible with any future effort levels Claude Code may add |

## Configuration

```json
{
  "display": {
    "showEffortLevel": true
  }
}
```

## Visual Examples

```
[Opus 4.6 ●max] ████░░░░░░ 45%     ← effort max
[Sonnet 4.6 ◑high] ██░░░░░░░░ 20%   ← effort high  
[Opus 4.6] ████░░░░░░ 45%           ← effort not detected (graceful)
```

## Test Plan

- [x] All 399 existing tests pass (0 failures)
- [x] 8 new effort tests pass (stdin priority, all levels, normalization, unknown levels)
- [x] `npm run build` succeeds
- [x] Manual test: compact layout shows `[Model ●max]`
- [x] Manual test: expanded layout shows `[Model ●max]`
- [x] Manual test: unknown level shows plain text (no symbol)
- [ ] Verify with real Claude Code session (requires `showEffortLevel: true` in config)

Closes #201
[...]

PR #491, opened 2026-04-24
fix(effort): tolerate object schema from Claude Code 2.1.115+

zelinewang:
## Problem

Since Claude Code 2.1.115+, the statusline crashes on every tick with:

```
[claude-hud] Error: level.toLowerCase is not a function
```

when a user has `display.showEffortLevel: true` in their config. The crash
only appears after upgrading Claude Code from 2.1.114 (or earlier) to
2.1.115+.

## Root cause

The original effort support (#471) was future-proofed for the case where
Claude Code would eventually expose the effort level via stdin JSON — but
because no schema existed yet, the type was guessed as a bare string:

```ts
effort?: string | null;
```

Claude Code 2.1.115+ actually added the field, but as an object:

```json
"effort": { "level": "max" },
"thinking": { "enabled": true },
"fast_mode": false
```

`resolveEffortLevel` then passes the whole object into `formatEffort`, which
invokes `level.toLowerCase()` on what is actually a `{ level: "max" }`
object — instant `TypeError`.

Evidence — live stdin dump from Claude Code 2.1.119 on macOS, Opus 4.7:

```json
{
  "session_id": "…",
  "version": "2.1.119",
  "model": { "id": "claude-opus-4-7", "display_name": "Opus 4.7" },
  "effort": { "level": "max" },
  "thinking": { "enabled": true },
  "fast_mode": false,
  …
}
```

## Fix

Accept both shapes (object with `level` string, or bare string) through a
tiny `extractEffortString` helper, and let any other truthy input fall
through to the existing `--effort` parent-process fallback rather than
throwing.

```ts
export type StdinEffortInput = string | StdinEffort | null | undefined;

export interface StdinEffort {
  level?: string | null;
  [key: string]: unknown;  // tolerate future additions like `budget`
}
```

The index signature keeps the type permissive — if Claude Code later adds
`effort.budget` or similar, no more breaking changes here.

## Tests

Added 10 tests in `tests/effort.test.js`:

- **Object schema (Claude Code 2.1.115+ shape)**
  - extracts `{ level: "max" }` → renders `●`
  - normalises uppercase level
  - handles all known levels inside objects
  - tolerates extra fields for forward-compat
  - falls through on missing / null / non-string `level`
- **Defensive handling (no crash)**
  - numeric / boolean / array inputs fall through rather than throwing

Result:

```
baseline: 486 pass, 0 fail
after   : 496 pass, 0 fail  (no regressions)
```

End-to-end verified by feeding the captured stdin payload to a locally-built
`dist/index.js`:

```
$ cat /tmp/hud-stdin-dump.json | node dist/index.js
[Opus 4.7 ●max] ██░░░░░░░░ 15% | zane | …
```

Exit code 0, effort bracket renders correctly.

## Notes

- `src/` only per CONTRIBUTING.md (CI rebuilds `dist/` on merge).
- Preserves backwards compatibility with the original PR #471 string shape
  in case any older Claude Code version still sends it or a user injects
  effort via another path.
- The `--effort` parent-process fallback path is unchanged and still acts
  as a safety net when stdin gives us nothing usable.

### DIFF
--- a/src/effort.ts
+++ b/src/effort.ts
@@ -31,11 +31,12 @@ export type StdinEffortInput = string | StdinEffort | null | undefined;
 /**
  * Resolve the current session's effort level.
  *
- * Priority:
- * 1. stdin.effort as object `{ level: string }` — Claude Code 2.1.115+
- * 2. stdin.effort as bare string — original PR #471 future-proofed path
- * 3. Parent process CLI args — `--effort` flag captured from ppid
- * 4. null
+ * Resolution order (matches `extractEffortString` below):
+ * 1. stdin.effort as non-empty string — original PR #471 future-proofed path.
+ * 2. stdin.effort as object with string `level` — Claude Code 2.1.115+ schema
+ *    (e.g., `{ "level": "max" }`).
+ * 3. Parent process CLI args — `--effort` flag captured from ppid.
+ * 4. null.
  *
  * Non-matching inputs (numbers, booleans, arrays, objects without a string
  * `level`) fall through to step 3 rather than crashing.
--- a/src/effort.ts
+++ b/src/effort.ts
@@ -14,17 +14,38 @@ const KNOWN_SYMBOLS: Record<string, string> = {
 };
 
 /**
+ * Shape of the effort field in Claude Code stdin JSON.
+ *
+ * Historically absent; Claude Code 2.1.115+ sends an object with a string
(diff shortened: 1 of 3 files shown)


jarrodwatts (approved):
Security, quality, and maintainability review cleared. Local full suite passed in an isolated worktree (496 pass, 1 skipped), and GitHub CI is green.

PR #492, opened 2026-04-24
fix(context): skip cache restore after /compact so the bar updates

Coiggahou2002:
## Summary

After running `/compact` (manual or auto), the context percentage bar stays at the pre-compact value instead of dropping. The session-scoped fallback cache added in #460 can't tell a legitimate post-`/compact` zero frame apart from the Claude Code "suspicious zero" glitch it was designed to paper over — both land on the exact same stdin shape.

## Root cause

The statusline's `context_window.used_percentage` is computed from the most recent assistant message's `usage` field. On `/compact`, Claude Code does two things:

1. **Zeroes out the `usage` field on every pre-compact assistant message.** So the percentage derived from "last assistant message with usage" becomes 0.
2. **Appends a `{type:"system", subtype:"compact_boundary", compactMetadata:{trigger, preTokens, postTokens, durationMs}}` entry** to the transcript.

Meanwhile `context_window.total_input_tokens` is cumulative over the entire session and is **not** reset by `/compact`.

The post-compact stdin (before the next assistant response lands) therefore looks like:

```json
{
  "context_window": {
    "context_window_size": 200000,
    "total_input_tokens": 500000,
    "used_percentage": 0,
    "remaining_percentage": 100,
    "current_usage": {
      "input_tokens": 0, "output_tokens": 0,
      "cache_creation_input_tokens": 0, "cache_read_input_tokens": 0
    }
  }
}
```

This matches `isSuspiciousZero` exactly — `context_window_size > 0`, `total_input_tokens > context_window_size`, `used_percentage === 0`, and `current_usage` all zero — so the cache restores the pre-compact snapshot. Any session that's been going long enough for `total_input_tokens` to cross `context_window_size` is affected, which happens after just a few turns.

## Fix

Use the `compact_boundary` transcript entry as the authoritative signal that a legitimate reset happened.

- `parseTranscript` now records the timestamp of the most recent `compact_boundary` and its `compactMetadata.postTokens` on `TranscriptData`.
- `applyContextWindowFallback` accepts a `CompactHint` and, when `isSuspiciousZero` fires:
  - If the last `compact_boundary` is newer than the cache's `saved_at` → treat the zero frame as a real post-compact reset, **do not** restore the stale snapshot.
  - If `compactMetadata.postTokens` is available → synthesize an accurate transition-window percent (`postTokens / context_window_size`) so the bar reflects the real post-compact context before the next assistant response arrives.
  - Otherwise (no hint, or boundary older than `saved_at`) → behave exactly as before, preserving #460's glitch protection.

The transcript cache version is bumped to 3 so existing caches re-parse once to pick up the new fields.

## Reproduction

<details><summary>Before/after script</summary>

```js
import { mkdtemp, rm } from 'node:fs/promises';
import { tmpdir } from 'node:os';
import path from 'node:path';
import { applyContextWindowFallback } from './dist/context-cache.js';
import { getContextPercent } from './dist/stdin.js';

const tempHome = await mkdtemp(path.join(tmpdir(), 'claude-hud-repro-'));
const transcriptPath = '/tmp/session.jsonl';

// pre-compact frame — cache captures 85%
const pre = {
  transcript_path: transcriptPath,
  context_window: {
    context_window_size: 200000, total_input_tokens: 500000,
    used_percentage: 85, remaining_percentage: 15,
    current_usage: { input_tokens: 150000, output_tokens: 100,
      cache_creation_input_tokens: 10000, cache_read_input_tokens: 20000 },
  },
};
applyContextWindowFallback(pre, { homeDir: () => tempHome, now: () => 1000, random: () => 1 });
console.log('pre-compact:', getContextPercent(pre) + '%');

// post-compact frame — zero usage, cumulative untouched
const post = {
  transcript_path: transcriptPath,
  context_window: {
    context_window_size: 200000, total_input_tokens: 500000,
    used_percentage: 0, remaining_percentage: 100,
    current_usage: { input_tokens: 0, output_tokens: 0,
      cache_creation_input_tokens: 0, cache_read_input_tokens: 0 },
  },
};
applyContextWindowFallback(post, { homeDir: () => tempHome, now: () => 2000, random: () => 1 },
  undefined, { lastCompactBoundaryAt: new Date(1500), lastCompactPostTokens: 7679 });
console.log('post-compact:', getContextPercent(post) + '%');

await rm(tempHome, { recursive: true, force: true });
```

- **Before**: `pre-compact: 85%` / `post-compact: 85%` (stuck)
- **After**:  `pre-compact: 85%` / `post-compact: 4%` (7679 / 200000 ≈ 3.84%)
</details>

## Test plan

- [x] `npm test` — 475 passing, 1 skipped (470 existing + 6 new)
- [x] Repro script shows the bar drops from 85% to 0% (or `postTokens/size` when available) after compact
- [x] Glitch path regression — when no `compact_boundary` is present, or it predates `saved_at`, the original #460 cache restore still fires

New test coverage:

- `parseTranscript` picks up the most recent `compact_boundary` + `postTokens`
- `parseTranscript` ignores malformed / wrong-subtype entries
- `applyContextWindowFallback` skips restore when boundary newer than `saved_at`
- Uses `postTokens` for the transition percent when provided
- Still restores cache when no hint is provided (glitch path preserved)
- Still restores cache when boundary is older than `saved_at` (glitch path preserved)

Only `src/` and `tests/` are modified, per CONTRIBUTING.md.


jarrodwatts (approved):
Security review cleared. I added maintainer regression coverage for main() compact metadata wiring and repeated post-compact cache refresh, merged current main into the contributor branch, and reran the full suite locally: 504 pass, 1 skipped. GitHub reports no checks for this fork branch, so local full-suite verification is the merge gate here.

PR #496, opened 2026-04-24
fix(speed-tracker): scope cache per session to prevent cross-session contamination

### LINKED ISSUE #495: Speed tracker cache is global, causing cross-session contamination
## Summary

`src/speed-tracker.ts` uses a single global cache file (`<plugin-dir>/.speed-cache.json`) instead of a per-session file. When two Claude Code sessions run concurrently, they read/write the same cache and contaminate each other's speed readings.

All other caches in the project correctly key by a hash of the transcript path (see `context-cache.ts` and the transcript cache). `speed-tracker.ts` is the only exception.

## Location

https://github.com/jarrodwatts/claude-hud/blob/main/src/speed-tracker.ts#L29-L31

```ts
function getCachePath(homeDir: string): string {
  return path.join(getHudPluginDir(homeDir), '.speed-cache.json');
}
```

## Reproduction

1. Open two Claude Code terminals simultaneously (Session A and Session B).
2. Let Session A stream a long response — `output_tokens` rises quickly, `.speed-cache.json` is written with A's counter.
3. Session B's statusline refresh (event-driven or via `refreshInterval`) reads the cache and uses A's `outputTokens` as its `previous`.
4. Because B's own `stdin.context_window.current_usage.output_tokens` is session-scoped (a different counter from A), the comparison is meaningless.

## Observed impact

- When B's counter happens to be larger than A's cached value and the write is within the 2s `SPEED_WINDOW_MS`, Session B's HUD shows a bogus tok/s reading that was actually caused by Session A's stream.
- When B's counter is smaller, B resets the cache to B's value, which then poisons A's next tick.

## Root cause

`output_tokens` in `stdin.context_window.current_usage` is a session-local counter. Comparing the current frame's value against a previous value written by a different session is a category error. The fix is to key the cache by transcript path, the same way `context-cache.ts` already does.

## Suggested fix

Key the cache file by a sha256 of the resolved `transcript_path` (matching `context-cache.ts`), and thread `transcript_path` through `getOutputSpeed`:

```ts
function getCachePath(homeDir: string, transcriptPath: string): string {
  const hash = createHash('sha256').update(path.resolve(transcriptPath)).digest('hex');
  return path.join(getHudPluginDir(homeDir), 'speed-cache', `${hash}.json`);
}
```

An opportunistic sweep (like `sweepCacheDir` in `context-cache.ts`) would keep the new directory bounded.

Happy to send a PR if this is the preferred shape.

## Environment

- Plugin version: 0.1.0
- Platform: macOS (darwin 25.4.0)
- Runtime: bun

YoYoSenRan:
Fixes #495

## Problem

`src/speed-tracker.ts` stored its cache in a single global file (`<plugin-dir>/.speed-cache.json`). Every concurrent Claude Code session shared that file. Because `output_tokens` in `stdin.context_window.current_usage` is a session-local counter, comparing the current frame against a previous value written by a different session produced:

- bogus `tok/s` readings on idle terminals whenever another terminal was actively streaming, and
- a poisoned cache for the streaming terminal as soon as the idle session reset the file to its own smaller value.

Every other cache in the project already isolates by a hash of the transcript path (see `context-cache.ts` and the transcript cache) — the speed cache was the only outlier.

## Fix

- Key the speed cache by `sha256(path.resolve(transcript_path))`, matching the `context-cache.ts` approach. Cache lives at `<plugin-dir>/speed-cache/<hash>.json`.
- Thread `transcript_path` through `getOutputSpeed`. When it is missing we cannot safely isolate cache entries, so we skip speed tracking rather than fall back to a cross-session file.
- Opportunistically remove the legacy global `.speed-cache.json` once on first call, so upgraded installs don't leave a dead file behind.

## Tests

- Added `getOutputSpeed returns null when transcript_path is missing`.
- Added `getOutputSpeed isolates cache across concurrent sessions`, which reproduces the original bug by interleaving two sessions and asserting the idle session's first tick seeds a fresh cache while the active session continues to report stable speeds.
- Updated existing tests and the `withDeterministicSpeedCache` render helper to seed the new per-session cache path and thread the transcript path through `stdin`.
- \`npm ci && npm run build && npm test\` → 506 passing, 0 failing, 1 skipped (pre-existing).

## Notes

- Only \`src/\` and \`tests/\` are modified per \`CONTRIBUTING.md\`; \`dist/\` is left for CI.
- The cache layout change means the legacy file is abandoned — the cleanup step removes it so it doesn't linger on disk.

### DIFF
--- a/src/speed-tracker.ts
+++ b/src/speed-tracker.ts
@@ -1,6 +1,7 @@
 import * as fs from 'node:fs';
 import * as path from 'node:path';
 import * as os from 'node:os';
+import { createHash } from 'node:crypto';
 import type { StdinData } from './types.js';
 import { getHudPluginDir } from './claude-config-dir.js';
 
@@ -10,6 +11,9 @@ const SPEED_WINDOW_MS = 2000;
 // spurious multi-thousand tok/s readings (see #481). Require at least
 // half a second of elapsed time before reporting a speed.
 const MIN_DELTA_MS = 500;
+
+const CACHE_DIRNAME = 'speed-cache';
+const LEGACY_CACHE_FILENAME = '.speed-cache.json';
 
 interface SpeedCache {
   outputTokens: number;
@@ -26,13 +30,19 @@ const defaultDeps: SpeedTrackerDeps = {
   now: () => Date.now(),
 };
 
-function getCachePath(homeDir: string): string {
-  return path.join(getHudPluginDir(homeDir), '.speed-cache.json');
+// Scope the cache by a sha256 of the resolved transcript path so that
+// concurrent Claude Code sessions never share or overwrite each other's
+// cached output-token counters. Sharing the cache across sessions
+// produced bogus speed readings on idle terminals whenever another
+// terminal was actively streaming (see #495).
+function getCachePath(homeDir: string, transcriptPath: string): string {
+  const hash = createHash('sha256').update(path.resolve(transcriptPath)).digest('hex');
+  return path.join(getHudPluginDir(homeDir), CACHE_DIRNAME, `${hash}.json`);
 }
 
-function readCache(homeDir: string): SpeedCache | null {
+function readCache(homeDir: string, transcriptPath: string): SpeedCache | null {
   try {
-    const cachePath = getCachePath(homeDir);
+    const cachePath = getCachePath(homeDir, transcriptPath);
     if (!fs.existsSync(cachePath)) return null;
     const content = fs.readFileSync(cachePath, 'utf8');
     const parsed = JSON.parse(content) as SpeedCache;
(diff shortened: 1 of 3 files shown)


jarrodwatts (approved):
Security, quality, regression, readability, and maintainability review passed.\n\nI reviewed the source/test-only diff, including cache path handling, filesystem writes/deletes, transcript_path handling, generated artifacts, package/script changes, and regression coverage. The cache key hashes the resolved transcript path and writes under the HUD plugin directory, so the cross-session contamination fix is scoped correctly.\n\nValidated in an isolated worktree with:\n\n- npm ci\n- npm test\n- npm audit --omit=dev --audit-level=moderate\n\nResult: 506 passing, 0 failing, 1 skipped; production audit reports 0 vulnerabilities.\n\nOne low-risk follow-up remains: the new speed-cache directory should eventually get bounded sweep/cap behavior like context-cache, to limit disk growth from many unique transcript paths. That does not block this bugfix.

PR #501, opened 2026-04-27
feat: add forceMaxWidth config override

x1han:
## Summary

Add an explicit `forceMaxWidth` config option so users can tell claude-hud to use `maxWidth` even when a terminal width is detected.

This fixes a real issue I hit on Linux when running claude-hud inside the Claude Code statusline environment: the HUD process saw a much smaller width than the actual visible terminal width, so long lines were unnecessarily truncated with `...`.

With this change:

- default behavior is unchanged
- `maxWidth` still works as a fallback by default
- users can opt in to `forceMaxWidth: true` when width detection is inaccurate

## Problem

On my Linux setup, Claude Code itself was displayed in a wide terminal, but the claude-hud subprocess only detected a narrow width. As a result, fields like:

- long branch names
- session token lines

were truncated even though there was enough visible horizontal space.

This was a width detection mismatch between the HUD subprocess and the actual terminal UI.

## Solution

Introduce:

```json
{
  "maxWidth": 300,
  "forceMaxWidth": true
}
```

When `forceMaxWidth` is enabled, claude-hud uses the configured `maxWidth` instead of the detected terminal width.

## Why this approach

- no magic-number heuristic
- no default behavior change
- fully opt-in
- easy to explain and document

## Test plan

- [x] Verify default behavior is unchanged when `forceMaxWidth` is not set
- [x] Verify detected width is still used by default
- [x] Verify `maxWidth` overrides detected width when `forceMaxWidth: true`
- [x] Verify config parsing accepts `forceMaxWidth: true`
- [x] Verify invalid `forceMaxWidth` values fall back to `false`
- [x] `npm test`

## Example

```json
{
  "lineLayout": "expanded",
  "maxWidth": 300,
  "forceMaxWidth": true
}
```

This resolved the truncation issue for me on Linux in Claude Code's statusline environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,7 @@ Chinese HUD labels are available as an explicit opt-in. English stays the defaul
 | `lineLayout` | string | `expanded` | Layout: `expanded` (multi-line) or `compact` (single line) |
 | `pathLevels` | 1-3 | 1 | Directory levels to show in project path |
 | `maxWidth` | number \| `null` | `null` | Optional fallback width used only when terminal width detection fails completely |
+| `forceMaxWidth` | boolean | false | Always use `maxWidth` when it is set, even if terminal width detection returns a smaller value |
 | `elementOrder` | string[] | `["project","context","usage","promptCache","memory","environment","tools","agents","todos"]` | Expanded-mode element order. Omit entries to hide them in expanded mode. |
 | `display.mergeGroups` | string[][] | `[["context","usage"]]` | Expanded-mode groups that should share a line when adjacent. Set `[]` to disable merged lines. |
 | `gitStatus.enabled` | boolean | true | Show git branch in HUD |
--- a/src/config.ts
+++ b/src/config.ts
@@ -71,6 +71,7 @@ export interface HudConfig {
   showSeparators: boolean;
   pathLevels: 1 | 2 | 3;
   maxWidth: number | null;
+  forceMaxWidth: boolean;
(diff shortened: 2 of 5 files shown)


jarrodwatts (changes requested):
Thanks for the focused report and implementation. I am holding this for changes before it can merge.

Required fixes:

- Remove all committed dist/ artifacts. This repo keeps PR review on source/tests/docs; CI builds dist after merge.
- Restore the existing showSessionName default coverage that was replaced in tests/config.test.js. Add forceMaxWidth tests as new coverage.
- Restore or preserve stderr-width fallback coverage in tests/render-width.test.js.
- Add coverage for forceMaxWidth with null/invalid maxWidth so the opt-in cannot accidentally erase detected width behavior.

This overlaps the active width-detection PRs (#510/#511), so keeping the diff source-only and narrowly tested matters here.

x1han:
Updated.

- removed committed dist artifacts from the PR
- restored the existing showSessionName default coverage
- restored stderr-width fallback coverage
- added forceMaxWidth coverage for null/invalid maxWidth

Full test suite passes.

jarrodwatts (approved):
Security, regression, and maintainability review passed after dist artifacts were removed and width fallback coverage was restored. The new forceMaxWidth option is opt-in and keeps default width behavior unchanged. Local validation: npm test passed with 518 passing, 1 skipped.

PR #502, opened 2026-04-27
feat(config): add barFilled and barEmpty color overrides for progress…

UnpxreTW:
## Summary

Add `colors.barFilled` and `colors.barEmpty` settings to let users customize the characters used in progress bars (context and usage). Defaults to the existing block characters (`█`/`░`) so behavior is unchanged without configuration.

### Output comparison

**Default (no config):**
```
[Opus]
Context ██████░░░░ 60% │ Usage █████░░░░░ 45%
```

**With `"barFilled": "●", "barEmpty": "○"`:**
```
[Opus]
Context ●●●●●●○○○○ 60% │ Usage ●●●●●○○○○○ 45%
```

**With `"barFilled": "▰", "barEmpty": "▱"`:**
```
[Opus]
Context ▰▰▰▰▰▰▱▱▱▱ 60% │ Usage ▰▰▰▰▰▱▱▱▱▱ 45%
```

### Config example

```json
{
  "colors": {
    "barFilled": "●",
    "barEmpty": "○"
  }
}
```

The settings live under `colors.*` rather than `display.*` because the bar rendering functions (`quotaBar` and `coloredBar`) already receive the `colors` object — this avoids changing any function signatures or call sites. While the settings are not color-related, they follow the existing pattern where `colors` acts as the visual style override namespace.

### Files changed

- `src/config.ts` — add `barFilled`/`barEmpty` to `HudColorOverrides`, defaults, and validation (single-character string)
- `src/render/colors.ts` — read custom chars from `colors` parameter in `quotaBar()` and `coloredBar()`
- `tests/render.test.js` — test custom bar characters render correctly in both usage and context bars
- `README.md` — document the two new settings in the colors options table

## Testing

- [x] `npm test`
- [x] `npm run test:coverage`

## Checklist


jarrodwatts (changes requested):
Thanks for the small, well-scoped config addition. I need changes before accepting this because these glyphs are emitted directly in the terminal statusline.

Required fixes:

- Reject control characters and escape characters for barFilled/barEmpty. The current value.length === 1 check accepts newline and ESC, which can corrupt rendering or inject terminal control behavior.
- Add config tests for invalid barFilled/barEmpty values.
- Validate intentionally by printable grapheme/cell width, so surrogate-pair glyphs and wide glyphs have explicit behavior instead of falling through JS string length semantics.

The default behavior is compatible. The blocker is validation of untrusted user-configured terminal output.

UnpxreTW:
Thanks for the review. Pushed three commits addressing all points:

## 1. Control character rejection
Extended the regex to also reject C1 control characters (0x80–0x9F) alongside C0 and DEL.

## 2. Grapheme validation
Replaced value.length === 1 with [...value].length === 1 (code point iteration) so surrogate-pair characters like emoji are correctly recognized as a single character instead of being silently rejected by JavaScript’s UTF-16 string length.

## 3. Wide glyph behavior
Wide characters (emoji, CJK, fullwidth forms) are accepted. Terminal cell width rendering varies across terminals and fonts, so a static Unicode range lookup cannot reliably determine actual display width. The README now notes that wide characters may affect bar alignment. Invalid values still fall back to the defaults.
Config validation tests added:
	•	Surrogate-pair emoji (accepted)
	•	CJK characters (accepted)
	•	C0 control characters (rejected)
	•	C1 control characters (rejected)
	•	Multi-character strings (rejected)
	•	Non-string types (rejected)

jarrodwatts:
Thanks for the follow-up validation work. This is close, but I am keeping it blocked for one remaining terminal-output safety issue: the current single-code-point check still allows invisible format/bidi characters such as U+202E into progress bar output.\n\nRequired next step: reject terminal-hostile format characters in barFilled/barEmpty, including bidi controls, zero-width joiners/non-joiners, variation selectors, and other invisible format/control code points. Please add focused config tests for those cases. After that, this should be small enough to re-review quickly.

UnpxreTW:
Thanks for catching the validation gaps. You were right that rejecting only C0/C1 control characters wasn’t sufficient.

Here’s what this update addresses:
Validation hardening

•	Switched to Intl.Segmenter for grapheme counting (replacing [...value].length), correctly handling compound emoji (e.g. 👨‍👩‍👧‍👦)

•	Replaced manual code point ranges with Unicode property escapes:
- [x] 	\p{Cc} — control characters
- [x] 	\p{Cf} — bidi controls (U+202E etc.), zero-width joiners/non-joiners, soft hyphen, word joiner, and all other invisible format characters
- [x] 	\p{Variation_Selector} — variation selectors (U+FE00–FE0F, U+E0100–E01FF)
- [x] 	\p{Zl} / \p{Zp} — line/paragraph separators (U+2028, U+2029 — not covered by \p{Cf})
- [x] 	\p{Cn} — noncharacters and unassigned code points

### Tests added

•	Dedicated tests for each rejection category: bidi controls, zero-width chars, variation selectors, invisible format chars, line/paragraph separators, noncharacters

•	Compound emoji still accepted

•	barFilled and barEmpty validated independently (one invalid doesn’t affect the other)

Let me know if there are any cases I’ve missed — happy to address them quickly.

jarrodwatts (approved):
Security, quality, and maintainability review passed after the maintainer patch. The validator now rejects any unsafe code point inside the accepted grapheme, generated dist artifacts were removed from the PR diff, and I ran npm test plus npm run test:coverage locally: 548 passing, 1 skipped; coverage completed successfully.

PR #505, opened 2026-04-28
feat(workspace): show /add-dir directories in statusline

dalamber:
## Summary

Surface the directories added via Claude Code's `/add-dir` command on the HUD's project line. Today the HUD reads `cwd` from stdin but ignores the `workspace` object entirely, so the rest of the session's working set is invisible.

Default (inline) — appended to the project name as dim, hyperlinked basenames with a `+` prefix per dir, no extra line:

```
[Opus] │ claude-hud +shared-utils +mobile-app git:(main*)
```

Optional standalone line via `display.addedDirsLayout: "line"` — uses an `Added dirs:` label and comma-separated names (no `+` prefix in this mode):

```
[Opus] │ claude-hud git:(main*)
Added dirs: shared-utils, mobile-app
```

Empty `added_dirs` array (the common case) renders nothing in either mode — no orphan label.

## Why this matters

`/add-dir` is the only way to bring multiple repos into a single Claude Code session, and it has no inverse: there is no `/remove-dir`. Dirs accumulate silently for the entire session, and once you've added more than one or two it's easy to lose track of what's actually in scope. For users who work across multiple repos daily — splitting one feature across an app and a shared library, debugging an integration that spans two services — this is a real ergonomic gap. The data is already on stdin; the HUD just wasn't reading it.

The information lives at `workspace.added_dirs` in the statusline stdin schema (Anthropic [statusline docs](https://code.claude.com/docs/en/statusline)): "Additional directories added via `/add-dir` or `--add-dir`. Empty array if none have been added."

## Implementation notes

- **`StdinData.workspace`** — added optional shape covering all four documented fields (`current_dir`, `project_dir`, `added_dirs`, `git_worktree`); only `added_dirs` is consumed today, the others are declared for type honesty without unused logic.
- **Two config keys**, both backwards-compatible:
  - `display.showAddedDirs` (boolean, default `true`)
  - `display.addedDirsLayout` (`'inline' | 'line'`, default `'inline'`)
- **Inline path** lives inside `project.ts` and shares the existing `branchOverflow: 'wrap'` behaviour — added dirs stay attached to the project name when the git block wraps.
- **Line path** is a new `addedDirs` `HudElement` in `DEFAULT_ELEMENT_ORDER`, slotted between `project` and `context`. The render returns `null` when the array is empty, so the line is fully suppressed (no orphan label).
- **Hyperlinks** use `file://` like the existing project name, so the dir basenames are clickable in terminals that support OSC-8.
- **Default chosen as `inline`** because most users never run `/add-dir`, and even those who do typically have 1–2 dirs — the standalone line is better suited to power users with many dirs.

## Testing

- [x] `npm test` — **510 pass, 1 skipped, 0 fail** (2 new integration tests)
- [x] Manual stdin test reproducing the docs schema with two added dirs — both inline and line layouts verified visually.
- [x] Empty `added_dirs: []` and missing `workspace` key — output identical to current behaviour (back-compat).
- [x] `showAddedDirs: false` — dirs hidden even when `added_dirs` is populated.
- [x] Live test inside Claude Code: `/add-dir ../some-folder` shows up within ~300ms.

New integration tests:
- `CLI renders added_dirs basenames on the project line` — asserts both basenames appear and that they're rendered after the project name.
- `CLI omits added dirs section when array is empty` — asserts no extra section appears when the array is empty.

## Checklist

- [x] Tests updated
- [x] Docs updated — new keys added to `README.md` config table
- [x] Per CONTRIBUTING.md, only `src/`, `tests/`, and `README.md` are touched — `dist/` left for CI

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,8 +170,8 @@ Chinese HUD labels are available as an explicit opt-in. English stays the defaul
 | `gitStatus.showFileStats` | boolean | false | Show file change counts `!M +A ✘D ?U` |
 | `gitStatus.branchOverflow` | `truncate` \| `wrap` | `truncate` | Keep current truncation behavior or let the git block wrap onto its own line boundary when possible |
 | `display.showModel` | boolean | true | Show model name `[Opus]` |
-| `display.showAddedDirs` | boolean | true | Show extra workspace directories from `/add-dir` (e.g. `+sparkle +lib-foo`); empty array renders nothing |
-| `display.addedDirsLayout` | `inline` \| `line` | `inline` | `inline` puts dirs next to the project name; `line` renders them on a separate `+dirs ...` line |
+| `display.showAddedDirs` | boolean | true | Show extra workspace directories from `/add-dir` (e.g. `+sparkle +lib-foo`); empty array renders nothing. In both layouts at most 5 dirs render (overflow shown as `+N more`) and basenames are truncated to 24 chars with `…` |
+| `display.addedDirsLayout` | `inline` \| `line` | `inline` | `inline` puts dirs next to the project name with a `+name` prefix per dir; `line` renders them on a separate `Added dirs: name1, name2` line (no `+` prefix, comma-separated) |
 | `display.showContextBar` | boolean | true | Show visual context bar `████░░░░░░` |
 | `display.contextValue` | `percent` \| `tokens` \| `remaining` \| `both` | `percent` | Context display format (`45%`, `45k/200k`, `55%` remaining, or `45% (45k/200k)`) |
 | `display.showConfigCounts` | boolean | false | Show CLAUDE.md, rules, MCPs, hooks counts |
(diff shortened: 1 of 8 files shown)


jarrodwatts (changes requested):
This is a useful feature and the direction fits the HUD: workspace.added_dirs is already available from stdin, and the empty default path stays quiet. I need a few guardrails before merge.

Required fixes:

- Treat malformed workspace.added_dirs defensively. Non-string entries should be ignored instead of reaching split/path rendering.
- Add tests for added_dirs: null, mixed non-string entries, line layout, showProject:false, and long directory names in narrow widths.
- Reconcile README wording with the rendered line layout. The body/docs describe the line mode differently than the code path that renders Added dirs:.
- Consider a small cap on rendered dir count or rendered basename length so oversized stdin cannot dominate the statusline.

Security review passed the current path/link handling; this is a quality and robustness hold.

dalamber:
> Required fixes:
> 
> * Treat malformed workspace.added_dirs defensively. Non-string entries should be ignored instead of reaching split/path rendering.
> * Add tests for added_dirs: null, mixed non-string entries, line layout, showProject:false, and long directory names in narrow widths.
> * Reconcile README wording with the rendered line layout. The body/docs describe the line mode differently than the code path that renders Added dirs:.
> * Consider a small cap on rendered dir count or rendered basename length so oversized stdin cannot dominate the statusline.
> 
> Security review passed the current path/link handling; this is a quality and robustness hold.

@jarrodwatts thanks for review, I have added 4 commits addressing your feedback.

jarrodwatts (approved):
Security, regression, and maintainability review passed after the added_dirs guardrails were added. The branch filters malformed stdin, strips control/bidi characters, uses file-only links, caps rendered output, and keeps the empty default path quiet. Local validation: npm test passed with 517 passing, 1 skipped.

PR #508, opened 2026-04-29
fix(context): treat nonzero token totals with zero usage as suspicious

lkangd:
## Summary

Fix a Claude Code data reporting glitch where `total_input_tokens` or `total_output_tokens` are non-zero, but all fields in `current_usage` are zero, causing `used_percentage` to incorrectly display as 0%.

## Problem

Claude Code sometimes sends malformed context window data like this:

```json
{
  "context_window": {
    "total_input_tokens": 978337,  // ← Non-zero accumulated tokens
    "total_output_tokens": 51018,  // ← Non-zero accumulated tokens
    "context_window_size": 1000000,
    "current_usage": {
      "input_tokens": 0,
      "output_tokens": 0,
      "cache_creation_input_tokens": 0,
      "cache_read_input_tokens": 0
    },
    "used_percentage": 0,  // ← Incorrectly shows 0% despite ~1M tokens used
    "remaining_percentage": 100
  }
}
```

Despite nearly 1 million tokens being consumed, the `current_usage` frame is entirely empty, causing the statusline to display 0% usage instead of the actual ~97% usage.

## Changes

- **types.ts**: Add `total_output_tokens` to `StdinData` interface
- **context-cache.ts**: Enhance `isSuspiciousZero()` detection logic:
  - **New condition**: Detect when accumulated totals (input or output) are non-zero but `used_percentage` is 0
  - **Existing condition**: Detect when total tokens exceed window size with zero usage
  - Update function documentation with detailed explanation of both trigger conditions
- **tests/context-cache.test.js**: Comprehensive test coverage for all `isSuspiciousZero()` branches:
  - New branch 1 logic: totalInputTokens/totalOutputTokens > 0 with usedPercentage === 0
  - Edge cases: zero tokens, zero window size, non-zero percentage
  - Updated existing test to align with new logic

## Solution

By checking both `total_input_tokens` and `total_output_tokens`, when either is non-zero but `used_percentage` is 0, the system now correctly identifies this as malformed data and triggers the cache fallback mechanism to use the last valid context snapshot.

## Test Plan

- [x] Code compiles and passes type checking
- [x] All 25 unit tests pass, including 6 new tests for branch coverage:
  - ✅ Detects suspicious zero when `totalInputTokens > 0` but `usedPercentage === 0`
  - ✅ Detects suspicious zero when `totalOutputTokens > 0` but `usedPercentage === 0`
  - ✅ Detects suspicious zero when both tokens > 0 but `usedPercentage === 0`
  - ✅ Does not treat as suspicious when all tokens and percentage are 0
  - ✅ Handles edge case when `contextWindowSize === 0`
  - ✅ Does not treat as suspicious when `used_percentage` is non-zero
- [x] Updated existing test to reflect new branch 1 logic behavior
- [ ] Verify statusline displays correct usage percentage when encountering malformed data
- [ ] Confirm legitimate zero-usage scenarios (session start) are unaffected
- [ ] Test with real Claude Code sessions exhibiting this glitch

## Technical Details

The fix adds an early detection path in `isSuspiciousZero()`:

```typescript
if ((totalInputTokens > 0 || totalOutputTokens > 0) && usedPercentage === 0) {
  return true;
}
```

This catches the glitch before the existing overflow detection logic, ensuring the statusline falls back to cached context data rather than displaying misleading 0% usage.

## Test Coverage

All conditional branches in `isSuspiciousZero()` are now fully tested:

1. **Branch 1** (new): Non-zero accumulated tokens with zero percentage → suspicious
2. **Branch 2**: No window size or tokens within bounds → not suspicious  
3. **Branch 3**: Non-zero percentage → not suspicious
4. **Branch 4**: All current_usage fields are zero → suspicious (existing logic)

Tests use helper functions (`makeSuspiciousFrame`, `makeHealthyFrame`) for consistency and maintainability.

### DIFF
--- a/src/context-cache.ts
+++ b/src/context-cache.ts
@@ -203,32 +203,25 @@ function isAllUsageZero(usage: ContextWindow["current_usage"]): boolean {
 
 /**
  * Returns true when context window data looks like a Claude Code reporting
- * glitch rather than a genuine zero-usage state. Two conditions trigger this:
+ * glitch rather than a genuine zero-usage state.
  *
- * 1. Accumulated token totals (input or output) are non-zero but
- *    `used_percentage` is 0 — the percentage hasn't caught up with reality.
- * 2. Total input tokens exceed the declared window size, `used_percentage`
- *    is still 0, AND every field in `current_usage` is also zero — the frame
- *    is entirely empty despite an overflow-level token count.
- *
- * Returns false when data is clearly valid: no window size, tokens within
- * bounds, or a non-zero `used_percentage` that confirms real usage.
+ * We only treat a zero-percent frame as suspicious when accumulated totals are
+ * non-zero and `current_usage` is still empty. If `current_usage` already shows
+ * non-zero token counters, keep the live frame instead of restoring stale cache.
  */
 function isSuspiciousZero(contextWindow: ContextWindow): boolean {
   const usedPercentage = contextWindow.used_percentage ?? 0;
-  const contextWindowSize = contextWindow.context_window_size ?? 0;
+  if (usedPercentage !== 0) {
+    return false;
+  }
+
+  if (!isAllUsageZero(contextWindow.current_usage)) {
+    return false;
+  }
+
   const totalInputTokens = contextWindow.total_input_tokens ?? 0;
   const totalOutputTokens = contextWindow.total_output_tokens ?? 0;
-  if ((totalInputTokens > 0 || totalOutputTokens > 0) && usedPercentage === 0) {
-    return true;
-  }
-  if (contextWindowSize <= 0 || totalInputTokens <= contextWindowSize) {
-    return false;
-  }
-  if ((contextWindow.used_percentage ?? 0) !== 0) {
-    return false;
-  }
-  return isAllUsageZero(contextWindow.current_usage);
+  return totalInputTokens > 0 || totalOutputTokens > 0;
 }
 
 /**
--- a/src/context-cache.ts
+++ b/src/context-cache.ts
@@ -202,12 +202,26 @@ function isAllUsageZero(usage: ContextWindow["current_usage"]): boolean {
 }
 
 /**
- * Detect the known Claude Code glitch where usage is reported as zero
- * despite a large accumulated input token count.
+ * Returns true when context window data looks like a Claude Code reporting
+ * glitch rather than a genuine zero-usage state. Two conditions trigger this:
+ *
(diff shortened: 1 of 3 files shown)


jarrodwatts (changes requested):
Thanks for isolating a real context-reporting failure mode. I am holding this because the new suspicious-zero condition is broader than the current guard and can preserve stale cached usage in legitimate reset-like frames.

Required fixes:

- Tighten the condition so nonzero total_input_tokens or total_output_tokens plus used_percentage: 0 is not enough by itself when current_usage contains usable data.
- Add coverage for nonzero totals with no cache, stale cache, nonzero current_usage, and a legitimate zero/reset frame.
- Keep the tests focused on behavioral risk. The current additions encode the broad fallback as desired behavior, which is the part that needs narrowing.

CI is green and the source surface is small, so this should be salvageable with a tighter predicate and edge-case tests.

lkangd:
> Thanks for isolating a real context-reporting failure mode. I am holding this because the new suspicious-zero condition is broader than the current guard and can preserve stale cached usage in legitimate reset-like frames.
> 
> Required fixes:
> 
> * Tighten the condition so nonzero total_input_tokens or total_output_tokens plus used_percentage: 0 is not enough by itself when current_usage contains usable data.
> * Add coverage for nonzero totals with no cache, stale cache, nonzero current_usage, and a legitimate zero/reset frame.
> * Keep the tests focused on behavioral risk. The current additions encode the broad fallback as desired behavior, which is the part that needs narrowing.
> 
> CI is green and the source surface is small, so this should be salvageable with a tighter predicate and edge-case tests.

Thanks for the review — I tightened the suspicious-zero fallback so we only restore cached context when `used_percentage` is 0, the accumulated totals are nonzero, and `current_usage` is still empty. If the zero-percent frame already carries usable `current_usage`, we now keep the live frame instead of overwriting it with stale cached data.

I also rewrote the tests around the behavioral risk you called out: nonzero totals with no cache, stale cache across compact/reset boundaries, zero-percent frames with nonzero `current_usage`, and legitimate zero/reset frames. The previous broad-fallback assertions have been removed.

jarrodwatts (approved):
Security, regression, and maintainability review passed after the suspicious-zero predicate was narrowed. The current branch preserves live frames with nonzero current_usage and covers reset/cache edge cases. Local validation: npm test passed with 511 passing, 1 skipped; GitHub Node 18/20 checks are green.

PR #509, opened 2026-04-29
fix(render): account for CJK ambiguous-width glyphs in width calculations

linziyanleo:
## Summary

- East Asian Ambiguous-width characters (`█` `░` `│` `⚠` `─` etc.) render as 2 cells in CJK terminals but were counted as 1 cell in width calculations
- This caused lines to overflow the terminal width without the code knowing, leading to broken wrapping and display issues in Claude Code's statusline UI
- Adds `src/render/width.ts` with `codePointCellWidth()` and `isCjkAmbiguousWide()` to detect CJK mode (when `language: 'zh'`) and account for double-width ambiguous chars
- Updates `visualLength()`, `makeSeparator()`, and label alignment to use the new CJK-aware width math

## Test plan

- [x] Added unit tests for `codePointCellWidth`, `isAmbiguousWideCodePoint`, `isCjkAmbiguousWide`
- [x] Added integration test: progress bars wrap correctly in CJK mode at narrow widths
- [x] Added integration test: separator `─` repeats halved in CJK mode so it doesn't overflow
- [x] Full test suite passes (511/511)

### DIFF
--- a/src/render/index.ts
+++ b/src/render/index.ts
@@ -17,6 +17,7 @@ import {
 } from './lines/index.js';
 import { dim, RESET } from './colors.js';
 import { getTerminalWidth, UNKNOWN_TERMINAL_WIDTH } from '../utils/terminal.js';
+import { codePointCellWidth, isCjkAmbiguousWide } from './width.js';
 
 // eslint-disable-next-line no-control-regex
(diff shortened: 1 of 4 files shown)


jarrodwatts (approved):
Validated locally with npm test in an isolated worktree: 511 passing, 0 failing, 1 skipped. Security review passed. This is a source/test-only renderer width fix, and the language-based CJK ambiguous-width heuristic is acceptable as a first pass.

Issue #506, opened 2026-04-29
Split `gitStatus.showFileStats` into separate inline-diff and file-list toggles


chy5301:
## Summary

Split the existing `gitStatus.showFileStats` boolean into two independent toggles so users can enable the inline line-diff (`[+44 -11]`) without rendering the per-file list line, or vice versa.

## Problem to Solve

`gitStatus.showFileStats: true` currently turns on two distinct rendering paths:

1. Inline line-diff inside the git segment of the project line: `git:(master* [+44 -11])` — `src/render/lines/project.ts:105-113`
2. A separate "git files" line listing up to 6 modified/added/deleted files — `renderGitFilesLine` at `src/render/lines/project.ts:191+`, called unconditionally from `src/render/index.ts:477`

There is no way to keep one and drop the other. For users on dirty branches (large refactors, design-doc work), the file-list line can dominate the status bar. The inline `[+N -M]` summary alone is often a sufficient "you have changes" cue, and per-file detail naturally belongs in `git status` / `git diff`.

## Proposed Solution

Introduce two new boolean keys (defaulting to the current behavior to stay backwards compatible):

- `gitStatus.showInlineDiff` — controls the `[+lines -lines]` inline annotation
- `gitStatus.showFileList` — controls the separate per-file list line

Keep `showFileStats` as a deprecated umbrella flag that still maps to both for one or two minor versions, then remove.

Implementation sketch:

- `renderProjectLine`: change the inline guard from `showFileStats` to `showInlineDiff`
- `renderGitFilesLine`: change the early-return guard from `showFileStats` to `showFileList`
- `mergeConfig`: when reading legacy `showFileStats`, if `showInlineDiff` / `showFileList` are not explicitly set, fall back to the legacy value

## Alternatives Considered

- Keep status quo and document a workaround → no real workaround exists short of patching `dist/`
- Auto-hide the file list when there are too many files → still all-or-nothing, doesn't address the core preference
- Combine into an enum (`'none' | 'inline' | 'list' | 'both'`) → less idiomatic vs the rest of the boolean-heavy display config

## Additional Context

Source pointers (v0.1.0):

- `src/render/lines/project.ts:58-169` — `renderProjectLine`
- `src/render/lines/project.ts:191-245` — `renderGitFilesLine`
- `src/render/index.ts:476-480` — unconditional call site for the file list
- `src/config.ts:75-83` — `gitStatus` config schema
[...]

PR #511, opened 2026-04-29
fix(render): enforce merge-group row budget and width-aware packing

linziyanleo:
## Summary

When the statusline runs in a narrow terminal, merge-group elements (context, usage, memory, environment, etc.) can overflow. This PR adds three related improvements:

- **Width-aware row packing**: `renderMergeSequence` now greedily packs as many elements as fit per row instead of placing each on its own line, then splits to a new row when the next element would exceed `terminalWidth`.
- **Compact density fallback**: When normal packing still exceeds the 2-row budget, elements re-render in `compact` mode (no progress bars, percentage-only) before trying again.
- **Consistent row budget enforcement**: Previously, the 2-row budget was gated behind `NARROW_MERGE_GROUP_WIDTH <= 160`. At widths 161-180, the check was bypassed, producing 3+ merge rows (4 total output lines), which Claude Code's statusline UI hid entirely. The budget now applies unconditionally.

### Depends on

This PR is based on #509 (CJK ambiguous-width fix). It can be reviewed independently but should be merged after #509.

## Test plan

- [x] New regression tests at 99, 160, 170, 180, and 80 columns verify the row budget holds
- [x] Greedy-packing test validates wide/mid/narrow split behavior
- [x] Stdin-columns piped mode tests verify width detection propagation
- [x] Full suite passes (521/521, 1 skipped)

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ Chinese HUD labels are available as an explicit opt-in. English stays the defaul
 | `pathLevels` | 1-3 | 1 | Directory levels to show in project path |
 | `maxWidth` | number \| `null` | `null` | Optional fallback width used only when terminal width detection fails completely |
 | `elementOrder` | string[] | `["project","context","usage","promptCache","memory","environment","tools","agents","todos"]` | Expanded-mode element order. Omit entries to hide them in expanded mode. |
-| `display.mergeGroups` | string[][] | `[["context","usage"]]` | Expanded-mode groups that should share a line when adjacent. Set `[]` to disable merged lines. |
+| `display.mergeGroups` | string[][] | `[["context","usage","promptCache","memory","environment"]]` | Expanded-mode groups that should share a line when adjacent. Set `[]` to disable merged lines. |
 | `gitStatus.enabled` | boolean | true | Show git branch in HUD |
 | `gitStatus.showDirty` | boolean | true | Show `*` for uncommitted changes |
 | `gitStatus.showAheadBehind` | boolean | false | Show `↑N ↓N` for ahead/behind remote |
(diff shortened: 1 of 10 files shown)


jarrodwatts (changes requested):
This needs to be split and cleaned up before review can continue.

Required fixes:

- Remove all dist/ artifacts from the PR. Generated output is a high-risk and hard-to-review surface in this repo.
- Rebase after #509 instead of bundling its CJK width commit into this branch.
- Split the row-budget/packing behavior into smaller PRs, or at least separate width math, packing, fallback density, and element-dropping policy.
- Do not change unknown-width behavior to force wrapping at a hardcoded width without a design decision. The existing no-width path intentionally lets the host render.
- Document and test any policy that drops opt-in elements such as memory, prompt cache, or environment to satisfy a row budget.
- Add coverage around gitStatus.showFileStats and issue #506 because the git files line is outside the merge-group budget.

The current diff is too broad for a safe maintainer review, and the generated files alone block merge.

jarrodwatts:
Rechecked after the latest update. This remains blocked as too broad for a safe merge. It still combines row packing, density fallback, element-dropping policy, width behavior, docs, and substantial tests in one PR, and it overlaps active width work.\n\nPlease split this into smaller source-only PRs. Start with one narrowly tested row-budget bugfix that preserves unknown-width behavior and does not drop opt-in elements. Packing strategy and fallback density can be reviewed separately after that.

jarrodwatts:
Thanks for the PR. I am closing this because the row-budget change is broad, stale against current main, and overlaps with the narrower wrapping work that has already landed. Please reopen as smaller focused changes if there are remaining concrete wrapping bugs after #571.

Issue #512, opened 2026-04-30
setup generates broken stty/awk wrapper that crashes on macOS when /dev/tty is available


gxgeek-n:
## Bug description

The `setup` command generates a `statusLine.command` wrapper that contains a bug in the terminal-width detection logic. When `/dev/tty` is available (normal for Claude Code desktop app on macOS), the wrapper script crashes before it ever launches Node.js, so the HUD never renders.

## Root cause

The generated bash wrapper includes:

```bash
cols=$(stty size </dev/tty 2>/dev/null | awk '{print }')
```

`stty size` outputs two numbers: `"rows columns"` (e.g., `"24 80"`). But `awk '{print }'` prints the **entire line** (both numbers with the space), not just the column count. The next line does arithmetic on this value:

```bash
export COLUMNS=$(( ${cols:-120} > 4 ? ${cols:-120} - 4 : 1 ))
```

When `cols="24 80"`, bash hits a syntax error:

```
bash: 24 80 > 4 ? 24 80 - 4 : 1 : syntax error in expression (error token is "80 > 4 ? 24 80 - 4 : 1 ")
```

The wrapper exits before reaching the `exec node ...` line.

## Why it's easy to miss

If `/dev/tty` is **not** available (e.g., in a non-TTY test environment, or certain CI contexts), `stty` fails immediately, `cols` stays empty, and the `${cols:-120}` fallback kicks in — so the bug is silently hidden. This is likely why it wasn't caught during development.

## Fix

Change `awk '{print }'` to `awk '{print $2}'` to correctly extract only the column count from `stty size` output.

Alternatively, since Claude Code already sets the `COLUMNS` environment variable, the `stty` probing could be replaced entirely with:

```bash
export COLUMNS=${COLUMNS:-120}
```

## Environment

- macOS (Apple Silicon)
- Claude Code v2.1.123
- claude-hud v0.1.0
- zsh

jarrodwatts:
Closing this as resolved in the current setup docs. The generated macOS/Linux command now extracts the column count with `awk '{print $2}'` and keeps the guarded fallback arithmetic, so the `awk '{print }'` crash described here is no longer present on `main`.\n\nThe broader setup-width simplification discussion can continue in #513, where version-safe plugin discovery and cache-layout compatibility still need to be preserved.

PR #513, opened 2026-04-30
fix: simplify COLUMNS detection in setup to avoid stty/awk fragility

gxgeek-n:
## Problem

The current setup.md instructs the LLM to generate a `statusLine.command` that probes terminal width with:

```bash
cols=$(stty size </dev/tty 2>/dev/null | awk '{print $2}')
```

This approach has two failure modes:

1. **LLM misinterpretation**: The shell quoting is extremely complex (`'"'"'` to nest single quotes). LLMs frequently drop the `$2`, generating `awk '{print }'` instead of `awk '{print $2}'`. When `/dev/tty` is available (macOS desktop app), `stty size` returns `"rows cols"`, `awk '{print }'` outputs both numbers, and the bash arithmetic `$(( 24 80 > 4 ? ... ))` hits a syntax error. The entire wrapper exits before reaching Node.js, and the HUD never appears. See #512.

2. **Unnecessary complexity**: Claude Code already sets the `COLUMNS` environment variable before invoking statusLine commands, making the `stty` probe redundant. If `/dev/tty` is unavailable, the fallback to 120 is used anyway.

## Fix

- Replace `stty` + `awk` terminal probing with `${COLUMNS:-120}`
- Replace the complex `awk/grep/sort` version-lookup pipeline with `ls -dt .../claude-hud/*/ | head -1` on macOS/Linux
- Update Windows + Git Bash instructions to also use `${COLUMNS:-120}`

## Test plan

- [x] Verified the simplified command works on macOS with Node.js
- [x] The command produces correct HUD output when stdin is provided
- [x] No regression in plugin path resolution

### DIFF
--- a/commands/setup.md
+++ b/commands/setup.md
@@ -150,32 +150,23 @@ This is a [Claude Code platform limitation](https://github.com/anthropics/claude
 
 5. Generate command (quotes around runtime path handle spaces):
 
-   The command exports `COLUMNS` so the HUD knows the real terminal width.
-   Claude Code pipes the subprocess stdout, so `process.stdout.columns` is
-   unavailable at runtime. `stty size </dev/tty` reads from the controlling
-   terminal. The `- 4` accounts for Claude Code's input area padding
-   (2 columns on each side).
+   Claude Code sets the `COLUMNS` environment variable before invoking the
+   statusLine command, so no terminal probing (`stty`/`tput`) is needed.
+   Fall back to 120 if the variable is somehow unset.
 
-   The grep pattern uses `[[:space:]]` rather than `\t` to match the tab
-   separator emitted by awk. GNU grep (BRE/ERE) does **not** interpret
-   `\t` as a tab character — it emits `warning: stray \ before t` and
-   treats the pattern as literal `t`, so the regex never matches the awk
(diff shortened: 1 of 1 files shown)


jarrodwatts:
Thanks for simplifying the setup command. I am keeping this blocked because the current branch changes two important behaviors without enough proof:\n\n- Plugin discovery switches from version-aware selection to mtime selection via ls -dt/head -1. That can select the wrong cached version.\n- The glob is narrowed to plugins/cache/claude-hud/claude-hud/*/, which drops compatibility with the broader marketplace cache shape the setup docs previously handled.\n\nPlease keep the safer COLUMNS simplification, but preserve version-safe plugin selection and the existing cache-layout compatibility. A dry-run or fixture-style setup test covering the generated path lookup would make this much easier to merge.

jarrodwatts:
Closing this now that the Windows setup overlap has been resolved by #532 and #538, and this branch still changes plugin version selection/cache-layout compatibility in ways that are unsafe to merge.\n\nA narrow follow-up would be welcome if it preserves the existing marketplace-aware, version-safe plugin lookup and only addresses the remaining COLUMNS simplification with tests or a dry-run fixture.

PR #515, opened 2026-04-30
fix: defer background agent completion to queue-operation for accurate timing and ◐ icon

gxgeek-n:
## Problem

Background agents (`run_in_background: true`) have two display bugs in the HUD:

1. **Wrong duration**: Agent ran 77s but HUD shows "4s"
2. **◐ icon never visible**: Agent flips from running to completed immediately

**Root cause**: In the parent transcript, `tool_result` for background agents is written as a launch acknowledgment (within ~4 seconds), not at actual completion. The HUD interprets this as the agent finishing.

## Fix

- Store `run_in_background` flag on the agent entry
- For **background agents**: skip the tool_result timestamp entirely; keep status `running` until the `queue-operation` entry provides the real completion time
- For **inline agents**: use tool_result timestamp as before (unchanged behavior)
- `queue-operation` entries already contain accurate `<task-id>` → `<tool-use-id>` mappings with real completion timestamps

## Result

| Agent | Before | After |
|-------|--------|-------|
| Long task X (77s actual) | ✓ 4s | ◐ during → ✓ 1m 17s |
| HUD running test (48s) | ✓ <1s | ◐ during → ✓ 48s |

## Test plan

- [x] Background agent shows yellow ◐ during execution
- [x] Background agent shows green ✓ with accurate duration after completion
- [x] Inline (foreground) agent still works correctly
- [x] Simultaneous foreground + background agents display properly

### DIFF
--- a/src/transcript.ts
+++ b/src/transcript.ts
@@ -317,11 +317,19 @@ export async function parseTranscript(transcriptPath: string): Promise<Transcrip
     // Return partial results on error
   }
 
-  // Override agent endTimes with accurate queue-operation timestamps where available
+  // Resolve agent completion: prefer queue-operation timestamps (accurate for
+  // background agents), fall back to tool_result timestamps (inline agents).
+  // Status is deferred so background agents show ◐ until they truly finish.
   for (const [toolUseId, endTime] of queueCompletionMap) {
     const agent = agentMap.get(toolUseId);
     if (agent) {
       agent.endTime = endTime;
+      agent.status = 'completed';
+    }
+  }
+  for (const agent of agentMap.values()) {
+    if (agent.status === 'running' && agent.endTime) {
+      agent.status = 'completed';
     }
   }
   result.tools = Array.from(toolMap.values()).slice(-20);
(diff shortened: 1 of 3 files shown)


jarrodwatts:
This targets a real background-agent timing problem, but I am keeping it blocked until the parser behavior is covered. The queue-operation parsing currently trusts transcript content enough to mark an agent completed, and there are no regression fixtures for that path.\n\nRequired next step: add transcript tests for a background Task launch acknowledgement, a still-running background agent before queue completion, queue completion with the mapped tool-use-id, an inline/foreground agent using tool_result as before, malformed queue-operation content, and an unrelated/forged tool-use-id that must not affect another agent. With that coverage, the change should be straightforward to re-review.

jarrodwatts (approved):
Approved after maintainer patching. Security review cleared the diff surface; added parser fixtures for background launch acknowledgments, mapped queue completions, foreground fallback, malformed queue content, and unrelated tool-use IDs. Local full suite passed: 515 passing, 1 skipped.

PR #517, opened 2026-04-30
feat(config): split showFileStats into showInlineDiff and showFileList (fixes #506)

### LINKED ISSUE #506: Split `gitStatus.showFileStats` into separate inline-diff and file-list toggles

kagura-agent:
## Summary

Split `gitStatus.showFileStats` into two independent boolean toggles:

- **`showInlineDiff`** — controls the `[+N -M]` line-diff annotation inside the project line
- **`showFileList`** — controls the separate per-file list line (`renderGitFilesLine`)

This lets users keep one and drop the other. For users on dirty branches (large refactors, etc.), the file-list line can dominate the status bar while the inline `[+N -M]` summary alone is a sufficient "you have changes" cue.

Closes #506

## Backward compatibility

`showFileStats` is preserved as a legacy umbrella flag. When `showInlineDiff`/`showFileList` are not explicitly set, `mergeConfig` maps `showFileStats` to both new keys. Explicit values always take priority.

## Changes

- **`src/config.ts`**: Add `showInlineDiff` and `showFileList` to the interface, defaults, and `mergeConfig` (with `showFileStats` fallback logic)
- **`src/render/lines/project.ts`**: Switch guards from `showFileStats` to `showInlineDiff` (line 55) and `showFileList` (line 138)
- **`tests/render.test.js`**: Update existing fixtures + add 4 new tests (independent toggle control, legacy fallback, explicit override)

## Testing

```
npm test
# 515 pass, 0 fail
```


kagura-agent:
Closing this PR — it looks like this repo doesn't actively review external contributions at the moment. Happy to re-open if there's interest. Thanks! 🌸

PR #523, opened 2026-05-07
fix(linux): use MemAvailable from /proc/meminfo instead of os.freemem()

### LINKED ISSUE #522: [Bug] Linux: memory usage inflated — uses MemFree instead of MemAvailable from /proc/meminfo
## Summary

On Linux, the memory usage indicator reads `os.freemem()` from Node.js, which maps to `MemFree` in `/proc/meminfo`. This excludes disk cache and buffers that the kernel can reclaim on demand, causing the displayed usage to be significantly inflated (e.g. 88% when actual available memory is ~43%).

## Steps to Reproduce

1. Run Claude HUD on Linux with `showMemoryUsage: true` and `lineLayout: expanded`.
2. Observe the memory percentage in the status line.
3. Compare against `free -h` — the "available" column will be much higher than what HUD implies.

## Expected Behavior

Memory usage reflects `MemAvailable` — truly usable memory including reclaimable cache. On a 30 GB system with 17 GB available, HUD should show ~43% used, not ~78–88%.

## Actual Behavior

HUD uses `os.freemem()` → `MemFree` (~6.8 GB on a 30 GB system), reporting ~88% used and misleading the user into thinking the system is under memory pressure.

## Environment

- OS: Linux (CachyOS, kernel 7.0.3) / zsh
- Node/Bun version: Bun 1.3.10
- Claude Code version: 2.1.128

## Logs or Screenshots

```
MemTotal:     31982940 kB
MemFree:       7105992 kB   ← what HUD uses (~78% used)
MemAvailable: 18290212 kB   ← what it should use (~43% used)
```

tiagor87:
## Summary

`os.freemem()` on Linux returns `MemFree` — only truly free RAM, excluding reclaimable disk cache and buffers. This inflates reported memory usage significantly (e.g. 88% when ~43% is actually used).

Add `readLinuxMemory` reader that reads `MemAvailable` from `/proc/meminfo` — the same metric used by `htop` and `free(1)`. Mirrors the existing `readMacOSMemory` pattern. Falls back to `readDefaultMemory` if `/proc/meminfo` is unavailable.

Closes #522

## Testing

- [ ] `npm test`
- [ ] `npm run test:coverage`

## Checklist

### DIFF
--- a/src/memory.ts
+++ b/src/memory.ts
@@ -22,6 +22,22 @@ export function parseVmStat(
   };
 }
 
+export function parseLinuxMeminfo(
+  output: string,
+): { totalBytes: number; freeBytes: number } | null {
+  const totalMatch = output.match(/^MemTotal:\s+(\d+)\s+kB/m);
+  const availMatch = output.match(/^MemAvailable:\s+(\d+)\s+kB/m);
+  if (!totalMatch || !availMatch) return null;
+
+  const totalBytes = Number(totalMatch[1]) * 1024;
+  const freeBytes = Number(availMatch[1]) * 1024;
+  if (!Number.isFinite(totalBytes) || !Number.isFinite(freeBytes)) {
+    return null;
+  }
+
+  return { totalBytes, freeBytes };
+}
+
 const readDefaultMemory: MemoryReader = () => ({
   totalBytes: os.totalmem(),
   freeBytes: os.freemem(),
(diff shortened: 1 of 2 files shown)


jarrodwatts (approved):
Security, quality, and maintainability review passed. I added focused parser coverage for MemAvailable handling and missing-field fallback, then ran npm test locally: 532 passing, 1 skipped.

PR #526, opened 2026-05-08
feat: add externalUsageCommand for auto-refreshing usage snapshots

zlj-zz:
## Summary

  Add `externalUsageCommand` to the display configuration, allowing HUD to automatically regenerate stale or missing external usage snapshots by executing a user-provided shell
  command.

  **Changes:**
  - `src/config.ts` — Add `externalUsageCommand: string` to `HudConfig.display` with validation and default `''`
  - `src/external-usage.ts` — Rewrite `getUsageFromExternalSnapshot` to:
    - Read the snapshot file first
    - If missing or stale **and** `externalUsageCommand` is configured, execute it synchronously (10s timeout)
    - Re-read the file after command execution before validating freshness
    - Extract parsing logic into `parseSnapshot`, `snapshotToUsageData`, `readSnapshotFile`, and `runExternalUsageCommand` helpers for clarity
  - `tests/stdin.test.js` — Clean up test imports after removing Kimi provider detection code

  This provides a generic mechanism for any external usage source (e.g., a standalone script querying the Kimi API) to feed data into HUD via `externalUsagePath` without requiring
   built-in provider-specific logic.

  ## Testing

  - [x] `npm test`
  - [ ] `npm run test:coverage`

  ## Checklist

  使用方式示例：

```json
  {
    "display": {
      "externalUsagePath": "/tmp/claude-hud-kimi-usage.json",
      "externalUsageCommand": "node /path/to/kimi-usage-fetcher.js",
      "externalUsageFreshnessMs": 60000
    }
  }
```

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -122,6 +122,7 @@ export interface HudConfig {
     sevenDayThreshold: number;
     environmentThreshold: number;
     externalUsagePath: string;
+    externalUsageCommand: string;
     externalUsageFreshnessMs: number;
     modelFormat: ModelFormatMode;
     modelOverride: string;
@@ -183,6 +184,7 @@ export const DEFAULT_CONFIG: HudConfig = {
     sevenDayThreshold: 80,
     environmentThreshold: 0,
     externalUsagePath: '',
+    externalUsageCommand: '',
     externalUsageFreshnessMs: 300000,
     modelFormat: 'full',
     modelOverride: '',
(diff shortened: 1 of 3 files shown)


jarrodwatts:
Thanks for the PR. I am closing this design as unsafe for the HUD runtime path.

The change runs `externalUsageCommand` with `spawnSync(..., shell: true)` from config when the snapshot is missing or stale. The statusline can be invoked every few hundred milliseconds, so a missing/stale snapshot or failing command can repeatedly block the HUD for up to 10 seconds. It also executes with the full process environment and the current workspace cwd, which creates credential-exposure and repo-hijack paths if a user copies a relative command into global config.

The safer direction is an out-of-process refresher that writes `externalUsagePath`, or a design with a strict cooldown, absolute command contract, explicit docs, and focused tests for success/failure/timeout/freshness behavior. The existing snapshot-only fallback stays in place.

PR #534, opened 2026-05-10
docs(setup): inject terminal width fallback into launch command

Gujiassh:
## Summary
- export `COLUMNS` from tmux pane width, then `tput cols`, then a safe default before launching the HUD plugin
- keep the generated setup command stable in shells where the status-line process starts without a populated width env
- limit the change to the setup template so existing runtime behavior stays untouched

## Validation
- inspected the generated shell command template in `commands/setup.md`
- verified the branch diff is limited to the setup command injection

### DIFF
--- a/commands/setup.md
+++ b/commands/setup.md
@@ -153,12 +153,12 @@ This is a [Claude Code platform limitation](https://github.com/anthropics/claude
 
    **When runtime is bun** - add `--env-file /dev/null` to prevent Bun from auto-loading project `.env` files:
    ```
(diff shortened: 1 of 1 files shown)


jarrodwatts:
Thanks for the PR. I am closing this because the branch is conflicted and based on older setup-command text. The current setup docs already use the version-aware marketplace cache glob and guarded `stty` width fallback, while this patch adds a separate `tmux`/`tput` path that would need a fresh review after rebase.

Width handling is being kept in the existing setup-width thread so we can preserve version-safe plugin discovery, cache-layout compatibility, and the `-4` Claude Code padding behavior together.

PR #550, opened 2026-05-14
feat: make bash command display length adapt to terminal width

li0near:
## Summary

Bash commands in the tools activity line were hardcoded to truncate at 30 characters, which often cuts off useful context (e.g., `git commit -m "Fix auth...` instead of the full command).

This PR:
- Makes truncation dynamic based on terminal width (`$COLUMNS`), so wider terminals show more of the command
- Adds a `display.toolTargetMaxLength` config option for explicit control:
  - `0` (default): auto-detect from terminal width
  - Any positive number: use as the max character length
- Updates the path truncation in `tools-line.ts` to also respect this config

## Config example

```json
{
  "display": {
    "toolTargetMaxLength": 80
  }
}
```

Or set to `0` (default) to auto-fit to terminal width.

## Test plan

- [x] All 576 existing tests pass
- [x] Updated the bash truncation test to verify terminal-width-based behavior
- [x] Build passes cleanly with `tsc`

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -107,6 +107,7 @@ export interface HudConfig {
     showResetLabel: boolean;
     usageCompact: boolean;
     showTools: boolean;
+    toolTargetMaxLength: number;
     showAgents: boolean;
     showTodos: boolean;
     showSessionName: boolean;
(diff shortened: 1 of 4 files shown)


jarrodwatts:
Thanks for the PR. This is superseded by #571, which merged configurable tool-name length and safer render-time wrapping. This branch also expands transcript-derived Bash targets using env-derived width and exposes longer untrusted command text on the tools line, which is too risky for the hot statusline path.

PR #558, opened 2026-05-18
feat: use line-style instead of block-style usage progress bars

adit-chandra:
## problem
- my team finds the block-style progress bars too visually heavy, and often distracting w.r.t. the other elements of the hud.

## solution, design, scope

- Switch progress bars from block/shade characters to cleaner line characters (`━` / `─`), similar to the progress bars that other tools (e.g. uv) use.
- Update docs, configuration examples, fixtures, and focused render/config tests.

## test plan

- `npm run build && node --test tests/config.test.js tests/render.test.js`
- `npm test`
- `git diff --check`

new visual:
Screenshot 2026-05-17 at 8 50 30 PM

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -87,7 +87,7 @@ src/
 
 ```
 [Opus] │ my-project git:(main*)
-Context █████░░░░░ 45% │ Usage ██░░░░░░░░ 25% (1h 30m / 5h)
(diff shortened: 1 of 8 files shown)


jarrodwatts:
Thanks for the PR. I am closing this because it changes the default visual style for every user. The default HUD bars need to stay stable. A future version could expose this as an opt-in bar character preset or document the existing colors.barFilled and colors.barEmpty settings.

PR #570, opened 2026-05-27
feat: write stdin rate_limits to optional JSON snapshot file

### LINKED ISSUE #539: Support writing stdin rate_limits to a local JSON snapshot file
Hi! First, thanks for building Claude HUD — it’s been really useful.

I’d like to request a small feature that seems very aligned with the current architecture:

## Request

When Claude Code provides `rate_limits` via stdin, optionally write that data to a local JSON snapshot file as well.

Claude HUD already has a read-side integration for external usage snapshots:

- `display.externalUsagePath`
- `display.externalUsageFreshnessMs`

So this would basically be the symmetric write-side of that idea:
- read `rate_limits` from official stdin
- write a small JSON snapshot locally
- no OAuth/API polling
- no undocumented endpoints
- no change to the current usage source of truth

## Why this would help

I’m currently building a small AI tool and I’d love to read Claude HUD’s usage/rate-limit data directly instead of reimplementing the same parsing / snapshot logic myself.

In other words, I don’t want to reinvent the wheel if Claude HUD is already the component closest to the official stdin data.

## Suggested behavior

Something like an optional config field could work well, for example:

- `display.externalUsageWritePath`
- or `display.writeExternalUsagePath`

And then when `rate_limits` is present on stdin, Claude HUD writes a JSON snapshot like:

```json
{
  "updated_at": "2026-05-12T08:00:00.000Z",
  "five_hour": {
    "used_percentage": 42,
    "resets_at": 1747036800
  },
  "seven_day": {
    "used_percentage": 18,
    "resets_at": 1747600000
  }
}
```

This seems like a very small, low-risk feature because:

- the data already exists in stdin
- Claude HUD already knows how to consume a similar snapshot format
- it would not reintroduce any API/OAuth/compliance concerns
- other local tools could reuse the same snapshot

## Important constraint

I’m **not** asking for Claude HUD to fetch usage from any separate API or credentials source.

Only:
- if official `rate_limits` exists in stdin,
- optionally persist it to disk.

I think this would make Claude HUD a nice local source of truth for the broader Claude Code tooling ecosystem.

Thanks for considering it.

mvanhorn:
## Summary
- Adds optional `display.externalUsageWritePath` config field (default unset = no write).
- On each render where stdin carries `rate_limits`, atomically writes a JSON snapshot to the configured path via tmp-file-then-rename.
- Schema mirrors the read-side `externalUsagePath` shape so the snapshot is the inverse: `{updated_at, five_hour: {used_percentage, resets_at}, seven_day: {used_percentage, resets_at}}`.
- Debounced to 30s + payload change to avoid writing on every statusline tick.
- No API polling, no credential reads, no fields beyond official stdin `rate_limits` keys.

## Tests
`pnpm test` -- 594 pass, 1 skipped. New test file `tests/external-usage.test.js` covers: happy path, no-op when path unset, no-op when stdin `rate_limits` missing, atomic write under rename failure, schema fidelity, debounce, and round-trip with the existing read path.

Closes #539.

AI was used for assistance.


mvanhorn:
Glad the optional rate_limits JSON snapshot path is in, thanks @jarrodwatts.

mvanhorn:
Thank you for merging the rate_limits JSON snapshot @jarrodwatts, downstream dashboards can read usage without scraping the HUD now.

Issue #569, opened 2026-05-26
Feature: Show invoked skill name with color differentiation in tools activity line


feibao:
## Summary

Show the **name of the most recently invoked Skill** in the tools activity line (or as a dedicated element), with distinct color styling — similar to how OMC HUD displays `skill:plan` or `skill:claude-hud:setup` in cyan.

## Motivation

Currently, when Claude Code invokes a Skill, the tools activity line only shows a generic count:

```
✓ Skill ×1 | ✓ Read ×7 | ✓ Bash ×4
```

This tells the user *that* a skill was called, but not *which* skill. In sessions with multiple skill invocations, or when a skill triggers a long-running operation, users lose situational awareness about what's happening.

Other tools like OMC HUD solve this by rendering the skill name directly:

```
skill:plan    skill:code-review    ultrawork
```

With color differentiation (e.g., cyan for skill names, magenta for active modes), users can instantly identify the current activity without scrolling through the conversation.

## Proposed Design

**Option A — Enhance the existing tools line:**

Replace the generic `Skill ×N` with the last invoked skill name:

```
◐ Skill: code-review | ✓ Read ×7 | ✓ Bash ×4
```

**Option B — New dedicated element (preferred):**

Add a `showSkills` toggle (default `false`, consistent with other activity toggles) that renders a separate line or inline badge:

```
skill: code-review(args)
```

With color styling:
- Skill name in **cyan** (distinct from tool names which are default color)
- Active modes (if applicable) in **magenta** or **bright magenta**
- Skill arguments truncated to ~15 chars

**Config:**

```json
{
  "display": {
    "showSkills": false
  },
  "colors": {
    "skill": "cyan"
  }
}
```

## Data Source

The skill name is available in the transcript JSONL via `tool_use.input.skill` when the tool name is `Skill`. The arguments are in `tool_use.input.args`. This is concrete metadata — no brittle text parsing required.

## Related

- #527 — Add showSkills / showMcp display elements (broader scope, focuses on active skill/MCP presence)
- #493 — Render skill name via tool_use.input.skill (closed, implemented the data source)
- #295 — Show agent/skill names instead of generic labels (closed)

This issue focuses specifically on the **visual presentation and color differentiation** of skill invocations, which is complementary to #527's element-level design.

## Use Case

[...]

Issue #527, opened 2026-05-08
Feature Request: Add showSkills / showMcp display elements


ancespio:
## Summary

Add `skills` and `mcp` as displayable HUD elements (alongside existing `tools`, `agents`, `todos`), so users can see at a glance which Skills and MCP servers are active in the current session.

## Motivation

Claude Code now supports both **Skills** (installable via `npx skills add`, including 19+ from Feishu CLI) and **MCP servers**. Currently there is no visual indicator in the HUD when these are in use — the only notification is a `<system-reminder>` block that appears inline in the conversation, which is easily missed during long sessions.

The `HudElement` type already supports 9 elements (`project`, `context`, `usage`, `promptCache`, `memory`, `environment`, `tools`, `agents`, `todos`). Extending this to include `skills` and `mcp` would give users better situational awareness.

## Proposed Design

Two new elements in `HudElement`:

```
| 'skills' | 'mcp'
```

Corresponding `display` toggles:

```json
"display": {
  "showSkills": false,   // default off, consistent with other activity toggles
  "showMcp": false       // default off
}
```

**Example rendering (Expanded layout):**
```
[Opus | Pro] │ my-project git:(main*)
Context ████░░░░░ 45% │ Usage ██░░░░░░░░ 25%
◐ skill: feishu-docs | ◐ skill: code-review
◐ mcp: linear (3 tools) | ◐ mcp: slack
```

**Example rendering (Compact layout):**
```
[Opus | Pro] ████░░░░░ 45% | my-project | ◐ skill: feishu-docs | ◐ mcp: linear
```

## Related

Issue #516 discusses simplifying MCP call counts — this is complementary. That issue is about how to count MCP interactions; this one is about surfacing skill/MCP presence in the status line at all.

## Alternatives Considered

- Relying on existing `showTools` to reflect MCP tool count (works but doesn't distinguish native vs MCP tools vs skills)
- `/skills` command for manual inspection (requires user to proactively check)

jarrodwatts:
Thanks, this is a useful backlog item. I labelled it as an enhancement.

The main design point is the data source. MCP server counts already come from config, while active skill detection needs a reliable source that does not depend on brittle parsing of inline reminders. A good PR should keep both displays default-off, avoid duplicating the existing tools/environment lines, and include transcript/config fixtures showing how skills and MCPs are detected.

mvanhorn:
[...]

Issue #499, opened 2026-04-27
Want to use ElementOrder+MergeGroup to customize the display order


LoopOuroboros:
## Summary
I want to use ElementOrder+MergeGroup to customize the display order

## Problem to Solve
Context + Useage + Token Speed in one line
sessionName + sessionTokens in one line，Not the last line

jarrodwatts:
Keeping this open as a layout customization request. The repo already has elementOrder and display.mergeGroups, so the useful next step is a focused proposal that shows which existing elements cannot be grouped today and what config shape would cover that without making the default expanded layout noisier. The broader row-budget work in #511 should be resolved separately before this gets implemented.

jarrodwatts:
Thanks for the layout customization request. `elementOrder` and `mergeGroups` already define the supported layout customization surface. The remaining need is a concrete config shape for cases that cannot be expressed today, plus tests that preserve the current default layout.

I’m closing this queue item for now. A focused proposal or PR with those details would be welcome.

PR #573, opened 2026-05-28
feat: add advisor model display (display.showAdvisor)

liuzhicanfly-dev:
## Summary

Surfaces the model configured via Claude Code's `/advisor` command so it sits inline on the project line:

```
[Opus 4.7 (1M context)] │ AthenaKB git:(main*) │ Advisor: Opus 4.7
Context ██░░░░░░░░ 19%
```

Default is **off** (`display.showAdvisor: false`) — fully opt-in.

## How it works

Claude Code stamps an `advisorModel` field onto every assistant record in the session transcript when `/advisor` is configured. The transcript parser already walks every line for tool / agent / todo state, so this just captures the latest non-empty `advisorModel` into the existing `TranscriptData` shape with no extra IO.

`renderProjectLine` then appends the prettified label (e.g. `claude-opus-4-7` → `Opus 4.7`) as another part on the same row, respecting the existing ` │ ` separator and ANSI handling. `display.advisorOverride` lets users force a custom label if they want a different display string.

## Changes

- `TranscriptData.advisorModel` field + parser capture (latest-wins, empty strings ignored)
- `TRANSCRIPT_CACHE_VERSION` bumped 5 → 6 to invalidate caches that predate the new field
- `display.showAdvisor` / `display.advisorOverride` config keys + validation
- `renderProjectLine` integration + `prettifyAdvisorId` helper (handles both canonical IDs and short aliases like `opus` / `sonnet` / `haiku`)
- i18n labels for `en` (`Advisor`) and `zh-Hans` (`顾问`)
- `commands/configure.md` Element Mapping table + Turn Off / Turn On flow entries

## Tests

15 new tests added under `tests/core.test.js`, `tests/config.test.js`, and `tests/render.test.js`:

- `prettifyAdvisorId` covers canonical IDs, short aliases, unknown formats, empty input
- `renderAdvisorLine` covers gating off, missing transcript data, transcript-driven labels, override-verbatim
- `parseTranscript` covers field extraction, missing-field undefined fallback, latest-wins, empty-string-ignored
- `mergeConfig` covers defaults, explicit override, length cap, non-boolean / non-string rejection

`npm test`: **628 / 630 pass** — the single failure is the pre-existing `countConfigs cache: miss on nested rules additions` test, which fails identically on `main` without any of these changes.

## Notes

Per `CONTRIBUTING.md`, `dist/` is intentionally **not** included in this PR — CI rebuilds it after merge.

## Test plan

- [x] \`npm ci && npm run build && npm test\` — clean (modulo the pre-existing unrelated failure noted above)
- [x] End-to-end smoke test: piped a real Claude Code stdin JSON + transcript through \`bun src/index.ts\` and confirmed \`Advisor: Opus 4.7\` renders inline on the project line
- [x] Verified \`display.showAdvisor: false\` (default) produces zero-overhead behavior (no extra parsing if the field is absent, just a one-line check per assistant record)


jarrodwatts (changes requested):
Thanks for the detailed implementation. I reviewed the PR body, discussion, diff, and local behavior in a temporary worktree. The feature shape is useful, and the default remains off, but this needs one security hardening pass before merge.

Required changes:

- Sanitize advisor display text before rendering. Both `ctx.transcript.advisorModel` and `display.advisorOverride` can currently reach the terminal output with ANSI/OSC/control/bidi sequences intact.
- Cap transcript-derived advisor text before caching or rendering so a malformed transcript value cannot become an oversized recurring statusline segment.
- Capture `advisorModel` only from assistant records, since the documented source is a top-level assistant field.
- Add tests covering CSI/OSC/control/bidi stripping, oversized values, and inline advisor placement in both compact and expanded rendering.
- Add README/README.zh config docs for `display.showAdvisor` and `display.advisorOverride`.

I ran the direct equivalent of build + tests here because `npm` is unavailable in this runtime: `./node_modules/.bin/tsc && node --test` in a temporary PR worktree. Result: 630 tests, 629 pass, 1 skip, 0 failures.

This PR is separate from #499, #527, and #569, so those should remain open after this lands.

liuzhicanfly-dev:
Thanks for the thorough review — all five points addressed in `e1cc2a3`. Re-requesting review.

**1. Sanitize advisor display text** — both `ctx.transcript.advisorModel` and `display.advisorOverride` now flow through the existing `sanitizeDisplayText` helper (already used for project paths / branch names) before rendering. ANSI ESC, OSC, BEL, DEL, C0/C1 controls, and bidi marks (LRM/RLM/LRO/RLO/PDF/isolates) are stripped. Defence-in-depth: the prettified result is re-sanitized in case any future change to `prettifyAdvisorId` reflects raw input bytes.

**2. Cap transcript-derived advisor text** — capped at parse time (`ADVISOR_MODEL_MAX_LEN = 64` in `transcript.ts`) before it reaches the JSON cache, and again at render time (`MAX_ADVISOR_DISPLAY_LEN = 64` in `advisor.ts`). The cache deserializer also re-applies the cap as belt-and-braces against legacy entries. `TRANSCRIPT_CACHE_VERSION` bumped 6 → 7 to invalidate caches that predate the cap.

**3. Capture `advisorModel` only from assistant records** — the parser now guards on `entry.type === 'assistant'` before reading the field, matching the documented schema.

**4. Tests** — 9 new tests in `tests/render.test.js` and `tests/core.test.js`:
- ANSI ESC + BEL + DEL bytes neutralised (asserts against ANSI-stripped output)
- Bidi marks (RLO/PDF) stripped from `advisorOverride`
- Oversize advisor text truncated at the display cap
- `advisorModel` capped at the transcript cap (separate test)
- Sanitize-collapses-to-empty input → renderer returns `null`
- Non-assistant records (`user`, `custom-title`, `system`) ignored
- Inline placement on the project line in expanded layout (`renderProjectLine`)
- Advisor absent when `showAdvisor: false` (project line)
- Inline placement on the session line in compact layout (`renderSessionLine`)

**5. README docs** — `display.showAdvisor` and `display.advisorOverride` rows added to the configuration tables in both `README.md` and `README.zh.md`.

`npm test`: **637 / 639 pass**, 1 skip, plus the same one pre-existing failure on `countConfigs cache: miss on nested rules additions` that I noted in the original PR — it fails identically on `main`. All new tests pass.

`dist/` is still intentionally untouched per `CONTRIBUTING.md`.

jarrodwatts (approved):
Fresh maintainer review after the hardening commit and the main-branch merge resolution: security, quality, regression, readability, and maintainability checks pass. I verified advisor text is sanitized and capped before rendering, transcript capture is assistant-only, generated dist remains out of the PR diff, and the branch passes the full local suite against current main: 653 pass, 1 skip, 0 failures.

Issue #581, opened 2026-05-29
Statusline wraps/flickers in terminals without a controlling TTY (e.g. Emdash): setup wrapper discards Claude Code's `$COLUMNS` and falls back to hardcoded 120


Fernandoaml:
### Summary
The `statusLine.command` generated by `/claude-hud:setup` computes the terminal width with `stty size </dev/tty`. When the statusline subprocess has **no controlling TTY** (which is the case under Claude Code — including, very visibly, inside the [Emdash](https://github.com/generalaction/emdash) terminal panes), `/dev/tty` fails and the command falls back to a hardcoded `120`. This **overwrites the correct `$COLUMNS` that Claude Code already exports**, so on any pane narrower than ~116 columns the multi-line HUD is truncated to the wrong (too-wide) width, wraps, and the statusline visibly flickers / disappears on each ~300ms refresh. On wide terminals the bug is invisible because 116 happens to fit.

### Environment
- Pop!_OS / COSMIC (Wayland), but OS-independent — reproduces in any host that gives Claude Code a narrow pane.
- claude-hud `0.1.0`, launched via `bun`.
- Most visible inside Emdash, whose panes are commonly 80×24.

### Symptom
HUD appears and disappears / flickers in narrow panes; renders fine in wide terminals.

### Root cause (verified)
The generated command is essentially:

```bash
cols=$(stty size </dev/tty 2>/dev/null | awk '{print $2}')
export COLUMNS=$(( ${cols:-120} > 4 ? ${cols:-120} - 4 : 1 ))
# ... locate plugin dir ...
exec bun ... src/index.ts
```

I instrumented the real statusline subprocess (logged its fds + env on every invocation). In **every** sample, across panes of different sizes:

- `stty size </dev/tty` → **`/dev/tty: No such device or address`** (the subprocess has no controlling TTY; stdin is a socket, stdout a pipe).
- **`$COLUMNS` is already set by Claude Code to the real pane width** — observed `80`, `121`, `139` matching the actual panes.

So `cols` is empty → `COLUMNS` is forced to `120 - 4 = 116`, **throwing away the correct value (e.g. 80)**. Since `src/render/index.ts` truncates each line to `getTerminalWidth()` (which reads `COLUMNS`), lines are truncated to 116 and overflow an 80-column pane → wrap → flicker.

Proof (simulating an 80-col pane):

```
[...]

Issue #555, opened 2026-05-17
Suggested Repair Log!


w409765910-glitch:
## Summary
# Windows setup generates statusLine command exceeding Claude Code's ~300ms render cadence — HUD silently fails

## Summary

On Windows, the `statusLine.command` generated by `/claude-hud:setup` takes **320–390ms per invocation** due to PowerShell wrapper overhead. Per `CLAUDE.md` ("The statusline is invoked every ~300ms by Claude Code"), this means **every call is slower than the call interval**. Claude Code appears to back off / circuit-break, and the HUD remains completely blank with no error visible to the user.

Calling `node.exe` directly bypassing the PowerShell wrapper drops invocation time to **~95ms** and the HUD renders normally. This suggests the Windows setup template should be revised.

## Environment

- OS: Windows 11 Pro 22631
- PowerShell: 5.1 (Windows PowerShell, the default)
- Node.js: v24.15.0 at `C:\Program Files\nodejs\node.exe`
- Claude Code: 2.1.143
- claude-hud: 0.1.0
- Install path: `C:\Users\<user>\.claude\plugins\cache\claude-hud\claude-hud\0.1.0\`

## Reproduction

1. Install claude-hud on Windows via the marketplace
2. Run `/claude-hud:setup`
3. Restart Claude Code
4. Observe: status line area above the input box is completely blank

## Root cause (timing data)

The setup-generated command:

```
powershell -NoProfile -NonInteractive -Command "$env:COLUMNS=120; $node=(Get-Command node -ErrorAction SilentlyContinue).Source; ...; $p=(Get-ChildItem ... 'plugins\cache\*\claude-hud\*' ...).FullName; ...; & $node $s"
```

Measured wall-clock time over 5 runs (stdin = stub session JSON, executed via `cmd /c "echo ... | <command>"`):

```
run 1: 340ms
run 2: 390ms
run 3: 372ms
run 4: 354ms
run 5: 320ms
```

Breakdown:
- PowerShell 5.1 startup: ~150–200ms
- `Get-ChildItem` + version sort over plugin cache: ~50–100ms
- `node.exe` cold start: ~100ms
- claude-hud script execution: ~30–50ms

**Total: 320–390ms, consistently above the ~300ms cadence documented in your CLAUDE.md.**

[...]

Issue #548, opened 2026-05-14
Windows setup can generate PowerShell statusLine that fails when Claude Code executes via Git Bash


wangwenxi-lw:
## Summary

On Windows, `claude-hud` was installed and the generated PowerShell-based `statusLine.command` worked when run manually, but Claude Code did not render the HUD. Debug logs showed Claude Code executing the configured statusLine and returning status `127`.

The root cause appears to be a shell mismatch: Claude Code detected/used Git Bash internally (`Using bash path: \"F:\Git\bin\bash.exe\"`), while the HUD setup had written a Windows `.cmd` / PowerShell-oriented statusLine command. The command was valid in PowerShell, but failed when invoked through Claude Code's statusline execution path.

## Environment

- OS: Windows 11
- Claude Code: `2.1.139`
- claude-hud: `0.0.10`
- Node: installed via nvm at `H:\nvm\nodejs\node.exe`
- Plugin cache path: `C:\Users\15503\.claude\plugins\cache\claude-hud\claude-hud\0.0.10`
- Claude Code debug log indicated Git Bash was being used: `Using bash path: \"F:\Git\bin\bash.exe\"`

## What happened

1. `claude-hud` was installed and present in `installed_plugins.json`.
2. The generated command in `~/.claude/settings.json` used a Windows PowerShell command to locate and run `dist/index.js`.
3. Running that command manually with sample statusLine stdin produced the expected HUD output.
4. The UI still did not show HUD output.
5. Added a wrapper script with logging and configured statusLine to call it.
6. Claude Code debug log showed:

```text
StatusLine [C:\Users\15503\.claude\claude-hud-statusline.cmd] completed with status 127
```

7. Switching the statusLine command to a Bash wrapper fixed the UI.

Working configuration:

```json
{
  "statusLine": {
    "type": "command",
    "command": "bash C:/Users/15503/.claude/claude-hud-statusline.sh"
  }
}
```

Wrapper:

```bash
#!/usr/bin/env bash
"/h/nvm/nodejs/node.exe" "$HOME/.claude/plugins/cache/claude-hud/claude-hud/0.0.10/dist/index.js"
```

## Expected behavior

The setup command should generate a statusLine command compatible with the shell Claude Code will actually use to execute `statusLine.command`, not only the user's apparent interactive shell/platform.

## Suggested improvement

- Detect whether Claude Code is using Git Bash on Windows for statusLine execution and generate a Bash-compatible command in that case.
[...]

PR #585, opened 2026-06-01
fix(setup): use dynamic runtime resolution instead of hardcoded paths

damndone:
## Problem

`/claude-hud:setup` currently hardcodes the absolute runtime path (e.g. `/Users/xxx/.nvm/versions/node/v24.16.0/bin/node`) into the generated `statusLine` command. When version managers like nvm, mise, or asdf switch the default Node.js version, this path becomes stale and the HUD silently breaks.

## Solution

Replace `{RUNTIME_PATH}` with `$(command -v node)` / `$(command -v bun)` in all generated statusLine commands. The runtime is resolved dynamically at each HUD refresh. PATH is inherited from Claude Code's parent shell, where version managers are already initialized, so it always finds the current version.

### Changes by platform

| Platform | Before | After |
|----------|--------|-------|
| macOS/Linux (bun) | `exec "{RUNTIME_PATH}" --env-file /dev/null` | `exec "$(command -v bun)" --env-file /dev/null` |
| macOS/Linux (node) | `exec "{RUNTIME_PATH}"` | `exec "$(command -v node)"` |
| Windows Git Bash | `exec "{RUNTIME_PATH}"` | `exec "$(command -v node)"` |
| Windows PowerShell | `& '{RUNTIME_PATH}'` | `& (Get-Command node).Source` |

### Other changes
- Simplified setup steps 2-4: no longer need to verify exact binary path with `ls -la`
- Updated troubleshooting docs to reflect dynamic resolution
- Fixed marketplace.json version (was still 0.0.12, now 0.1.1)

## Verification

Tested on macOS with nvm-managed Node.js v24.16.0:
```
$ bash -c '... exec "$(command -v node)" "${plugin_dir}dist/index.js"'
[claude-hud] Initializing...
```

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -6,7 +6,7 @@
   },
   "metadata": {
     "description": "Real-time statusline HUD for Claude Code - context health, tool activity, agent tracking, and todo progress",
-    "version": "0.0.12"
+    "version": "0.1.1"
   },
   "plugins": [
     {
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "claude-hud",
   "description": "Real-time statusline HUD for Claude Code - context health, tool activity, agent tracking, and todo progress",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "author": {
     "name": "Jarrod Watts",
     "url": "https://github.com/jarrodwatts"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,10 @@
 
 All notable changes to Claude HUD will be documented in this file.
 
-## [Unreleased]
+## [0.1.1] - 2026-06-01
 
 ### Fixed
+- `/claude-hud:setup` no longer hardcodes the absolute runtime path in the generated statusLine command. Uses `command -v` for dynamic runtime resolution, so HUD automatically adapts when version managers (nvm, mise, asdf) switch Node.js versions.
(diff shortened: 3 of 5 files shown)


jarrodwatts (changes requested):
Thanks for the setup work. I cannot merge this shape.

Security and reliability blockers:

- The generated statusline commands resolve `node` or `bun` through `PATH` on every refresh. A statusline runs repeatedly and should execute a setup-time validated runtime path or a tightly controlled wrapper. Runtime `PATH` lookup makes the command sensitive to version-manager state, project-local path injection, and GUI/minimal-shell environments.
- The PowerShell path has the same issue with `(Get-Command node).Source` at render time.
- The setup instructions are internally inconsistent: some helper/wrapper sections still refer to `__RUNTIME_PATH__` and `$runtimePathLiteral` while other sections moved to dynamic lookup.
- The PR includes release/version/marketplace churn that is outside this fix and lacks the corresponding release mechanics.
- It does not address the broader setup cluster cleanly: #581 needs `$COLUMNS` preservation, #555 needs the Windows PowerShell render-cost problem solved, and #548 needs Windows Git Bash execution compatibility.

A mergeable version should be a focused setup redesign: source/docs only, no release bump, no generated artifacts, a runtime path resolved and validated during setup or a safe generated wrapper, no per-refresh `PATH` trust, and coverage or documented manual verification for macOS/Linux, Windows PowerShell, and Windows Git Bash.

jarrodwatts:
Thanks for digging into the runtime-path setup issue. I moved the fix into maintainer-owned PR #600 so the launcher is generated once during setup and the per-refresh command stays small.

I’m closing this older branch because it is conflicted and the replacement covers the setup failure with a narrower diff. Focused follow-ups on setup portability are welcome.

PR #589, opened 2026-06-03
feat: add optional autoCompactWindow context denominator (#540)

### LINKED ISSUE #540: Context display should use autocompactWindow as denominator instead of full context window
## Current Behavior

- **HUD shows:** 7% context usage (based on 1M context window)
- **`/context` command shows:** 70.6k/200k tokens (35%)

As you can see in the attached screenshots:

**Screenshot 1 - HUD Display:**

Image

Shows context at 7% using the full 1M context window as the denominator.

**Screenshot 2 - `/context` Output:**

Image

Shows the actual usage at 70.6k/200k tokens (35%), where 200k is the effective `autoCompactWindow`.

## The Problem

When a user explicitly sets `autoCompactWindow: 200k` in their Claude Code configuration, they care about:
- How close they are to triggering auto-compression (compression happens at 200k)
- NOT how much of the theoretical 1M model capacity they're using

The 7% figure is misleading because it suggests plenty of room left, while in reality they're at 35% of the autocompact threshold.

## Expected Behavior

HUD should respect the `autoCompactWindow` setting and use it as the denominator when:
1. `autoCompactWindow` is explicitly configured
2. Display context usage relative to when auto-compression will trigger

Ideally showing something like: "Context 70.6k/200k (35%)" to match what `/context` displays.

## Questions

1. Is there currently a configuration option to change the context display denominator?
2. Would you be open to a feature where HUD respects `autoCompactWindow` if set?

## Environment

  - Claude: 2.1.137
  - Claude Code: Opus 4.7 (1M context)
  - claude-hud: Latest
  - autoCompactWindow: 200k (configured in Claude Code settings)

mvanhorn:
## Summary

Adds an optional `display.autoCompactWindow` config (default `null`). When set to a positive number (e.g. `200000`), the HUD computes the context percentage and token display against the configured auto-compact window instead of the full model context window. When unset/null, behavior is unchanged.

## Why this matters

Issue #540: with a 1M model window but an effective 200k auto-compact threshold, the HUD reported ~7% (70.6k/1M) while `/context` reported 35% (70.6k/200k). The full-window figure hides how close a session is to triggering auto-compaction.

Per @jarrodwatts's guidance on the issue, this preserves current defaults and uses an explicit HUD config rather than broad implicit reads of Claude settings. As @truffle-dev confirmed, Claude Code's statusline stdin does not expose an `autoCompactWindow` field, so the override has to short-circuit the native `used_percentage` path rather than feed through the existing fallback.

Behavior:
- `null` (default): today's path unchanged (prefer native `used_percentage`, fall back to `context_window_size`).
- positive number: ignore the native percentage and use `min(100, round(totalTokens / autoCompactWindow * 100))` in both `getContextPercent` and `getBufferedPercent`. The token / `both` displays use the same denominator so the bar, the percent, and the token figure all agree with `/context` (e.g. `35% (71k/200k)`).
- invalid values (`0`, negative, `NaN`, non-numeric) coerce to `null`.

## Testing

- `npm test` (build + `node --test`): 629 passing, 1 pre-existing skip.
- New coverage in `tests/stdin.test.js`: null-default regression, 200k denominator → 35%, over-window clamp to 100, config validation of invalid values, and native-percentage override.
- New render test in `tests/render.test.js`: compact `both` mode shows `35% (71k/200k)` not `7% (71k/1.0M)`.
- `dist/` is intentionally left for the existing build-dist CI step to regenerate post-merge.

Fixes #540

AI was used for assistance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,7 @@ Chinese HUD labels are available as an explicit opt-in. English stays the defaul
 | `display.addedDirsLayout` | `inline` \| `line` | `inline` | `inline` puts dirs next to the project name with a `+name` prefix per dir; `line` renders them on a separate `Added dirs: name1, name2` line (no `+` prefix, comma-separated) |
 | `display.showContextBar` | boolean | true | Show visual context bar `████░░░░░░` |
 | `display.contextValue` | `percent` \| `tokens` \| `remaining` \| `both` | `percent` | Context display format (`45%`, `45k/200k`, `55%` remaining, or `45% (45k/200k)`) |
+| `display.autoCompactWindow` | number \| `null` | `null` | When set to a positive number such as `200000`, compute the context percentage against this auto-compact window instead of the full model context window, matching the `/context` figure. Leave unset or `null` to preserve default full-window behavior. |
 | `display.showConfigCounts` | boolean | false | Show CLAUDE.md, rules, MCPs, hooks counts |
 | `display.showCost` | boolean | false | Show session cost using Claude Code's native `cost.total_cost_usd` when available, with a local estimate fallback for direct Anthropic sessions |
 | `display.showOutputStyle` | boolean | false | Show the active Claude Code `outputStyle` from settings files as `style: <name>` |
--- a/src/config.ts
+++ b/src/config.ts
@@ -444,10 +444,10 @@ function validateNonNegativeInteger(value: unknown, fallback: number): number {
 }
 
 function validateAutoCompactWindow(value: unknown): number | null {
-  if (typeof value !== 'number' || !Number.isFinite(value) || value <= 0) {
+  if (typeof value !== 'number' || !Number.isFinite(value) || !Number.isInteger(value) || value <= 0) {
     return null;
   }
-  return Math.floor(value);
+  return value;
 }
 
 function validateOptionalPath(value: unknown): string {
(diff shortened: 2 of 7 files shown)


jarrodwatts (approved):
Security, quality, regression, readability, and maintainability review passed. I added one small maintainer follow-up to require a positive integer for display.autoCompactWindow, then reran the full local suite and waited for Node 18/20 CI to pass.

truffle-dev:
The invalid-value coercion to null (0 / negative / NaN / non-numeric) is a nice addition. That's what keeps a typo in the config from silently flattening the bar to 0%. The compact-mode render assertion (`35% (71k/200k)` not `7% (71k/1.0M)`) reads exactly the way `/context` does, which is the whole point. Thanks for picking this up.

mvanhorn:
Much appreciated @jarrodwatts! Glad the autoCompactWindow denominator option fit the HUD design.

truffle-dev:
Nice landing — glad the override slotted in cleanly. Thanks for the careful credit in the body, @mvanhorn.

truffle-dev:
Thanks for picking up #540 and landing it. The shared-denominator path you took makes the bar, the percent, and the token figure all match `/context`, which is what keeps it from feeling like a flag.

mvanhorn:
Cheers @jarrodwatts, thanks for taking the autoCompactWindow denominator - context percentages should read truer for auto-compact users.

PR #595, opened 2026-06-07
feat: add showSkills / showMcp HUD display elements

mvanhorn:
## Summary

Adds `skills` and `mcp` as displayable HUD elements, requested in #527. When enabled via `display.showSkills` / `display.showMcp`, the HUD shows which Skills and MCP servers are active in the current session, alongside the existing `tools`, `agents`, and `todos` elements. Both default to off, so existing configs render exactly as before.

Skill and MCP activity is detected from the transcript (`src/transcript.ts`): Skill tool invocations populate the skills line, and `mcp__server__tool` calls are aggregated per server for the MCP line. Rendering follows the same overflow rules as the tools line (`src/render/skills-mcp-line.ts`), and the new elements participate in `elementOrder` like every other HudElement.

`mergeConfig` migration handles configs written before these fields existed (missing booleans fall back to defaults), and the README documents both options.

AI was used for assistance.

## Testing

- [x] `npm test` (659 tests, 658 pass, 1 skipped; includes new render + transcript cases for skills/MCP extraction, ordering, overflow, and config migration)
- [ ] `npm run test:coverage`

## Checklist

- [x] Tests updated or not needed
- [x] Docs updated if behavior changed

Fixes #527

PR #596, opened 2026-06-07
feat: add Skills line — detect manually-loaded skills, no double-display (re: #527, #595)

AvishayHirsh:
### Why

I started using the `showSkills` element and ran into two things I wanted to fix:

1. **It only listed skills the _model_ invoked** via the `Skill` tool. Skills I loaded **myself** by typing `/skill-name`, or that got pulled in automatically, never showed up — even though they were active in the session.
2. **With both the tools line and the skills line enabled, a skill showed up twice** — once on the Skills line and again as `✓ Skill` on the tools line.

I wanted the line to reflect *every* skill active in my session, however it loaded, and to show each skill exactly once. That's the motivation for this PR.

I saw **#595** already adds `showSkills`/`showMcp` from `Skill` tool calls (thanks @mvanhorn) — this isn't a replacement (it's skills-only and focused on the two gaps above). Glad to fold it into #595 or close if you'd prefer that approach.

### Demo
Screenshot 2026-06-07 at 19 47 54

_Screenshot: the `⚡ Skills:` line showing skills loaded different ways (model-invoked, user `/skill`, auto) — note the tools line directly above has **no** duplicate `Skill` entry._

### How it works

- **Detection:** alongside the `Skill` tool signal, it reads the `Base directory for this skill:` line the harness writes whenever a folder-backed skill loads. That line appears no matter how the skill was triggered — model, a user-typed `/skill`, auto-load, or sub-agent — so manual/auto loads finally show up. Both signals normalize to the bare skill name and dedupe, so a skill seen twice still lists once.
- **No duplicate display:** when `showSkills` is on, the `Skill` tool is dropped from the tools line, so each skill appears in exactly one place.

### Design notes from #527

- **Reliable source, not reminder-scraping:** the marker is matched only on `isMeta` user messages and is start-anchored (`^`), so it keys off the harness's actual skill-load record — not `<system-reminder>` text or prose.
- **Avoid duplicating existing lines (#527):** the tools-line suppression above.
- **Default-off:** `display.showSkills` defaults `false`; existing configs render unchanged.
- **Tests/fixtures:** transcript detection + anchoring (incl. a quoted-string negative case), two-signal dedup, cache round-trip, render cap/overflow, tools-line suppression, and config.

### Tradeoff

Catching user-invoked skills means reading that injected load marker — there's no structured tool call for them. I aimed for reliable rather than brittle, but it's a different data-source choice than #595's tool-call-only approach. Open to discussion.

Skills-only by design; #595's MCP work is complementary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,7 @@ Chinese HUD labels are available as an explicit opt-in. English stays the defaul
 | `display.toolsMaxVisible` | number | `4` | Maximum completed tools shown on the tools line. `0` means unlimited |
 | `display.showAgents` | boolean | false | Show agents activity line |
 | `display.showTodos` | boolean | false | Show todos progress line |
+| `display.showSkills` | boolean | false | Show a `⚡ Skills:` line listing skills used this session. Detected from `Skill` tool calls and from the `Base directory for this skill:` load marker (so user-invoked `/skill` and auto-loaded skills are included), deduped to bare names. When on, the `Skill` tool is omitted from the tools line so each skill appears once. |
 | `display.showSessionName` | boolean | false | Show session slug or custom title from `/rename` |
 | `display.showAdvisor` | boolean | false | Inline the model configured via Claude Code's `/advisor` on the project line, e.g. `Advisor: Opus 4.7`. Read from the `advisorModel` field that Claude Code stamps on each assistant transcript record; sanitised and capped at 64 chars before rendering |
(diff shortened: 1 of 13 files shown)


jarrodwatts:
Thanks for pushing the manual-skill detection idea forward. #595 is now merged with default-off Skills/MCP lines, activity-name sanitization, length caps, and Skill-tool suppression.

I’m closing this stacked branch to keep the merged surface focused. A smaller follow-up for manual skill marker detection, rebased on current main, would be welcome.

PR #611, opened 2026-06-12
feat: aggregate Skill display and support string content parsing

vector4wang:
- Parse <command-name> tags in string content to capture /skill invocations
- Track Skill running/completed state transitions via runningSkillIds queue
- Aggregate all completed Skills under single "Skill" key (✓ Skill ×N)
- Cap Skill entries at 50 to prevent unbounded growth in long sessions
- Unique Skill IDs via timestamp + counter to prevent collisions
- Preserve Skill entries from .slice(-20) truncation

All 673 tests passing.

## Testing

- [ ] `npm test`
- [ ] `npm run test:coverage`

## Checklist

### DIFF
--- a/src/render/tools-line.ts
+++ b/src/render/tools-line.ts
@@ -46,8 +46,10 @@ export function renderToolsLine(ctx: RenderContext): string | null {
 
   const toolCounts = new Map<string, number>();
   for (const tool of completedTools) {
-    const count = toolCounts.get(tool.name) ?? 0;
-    toolCounts.set(tool.name, count + 1);
+    // Aggregate all Skills under single "Skill" key
+    const key = tool.name === 'Skill' ? 'Skill' : tool.name;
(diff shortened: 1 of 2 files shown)


jarrodwatts:
Thanks for the PR. I’m closing this after maintainer review.

The current parser treats any `<command-name>/...` string as Skill activity, so built-in slash commands such as `/compact` can appear as `Skill` and be added to the skills list. It also changes the retained tool output from the last 20 tools to up to 50 synthetic Skill entries plus 20 other tools, and it changes transcript parser output without a cache-version bump or focused tests.

A safer follow-up should preserve the existing tool chronology and cap, avoid classifying built-in slash commands as skills, bump the transcript cache version when parsed output changes, and add transcript fixtures for string content, hostile or oversized command names, cache reuse, and `showTools`/`showSkills` interaction.

PR #647, opened 2026-06-29
refactor(i18n): let locales own layout via a tiny interpolate() helper

UnpxreTW:
## Summary

Layout that varies by language (spacing, affix position) is currently hardcoded in render code rather than owned by the locale. Two existing examples:

- `session-time.ts` hardcodes the relative-time suffix spacing behind a `getLanguage() === 'zh'` branch.
- `format-reset-time.ts` uses an empty-string key (`format.at: ""`) as a workaround so locales that bake the preposition elsewhere get a bare time.

This adds a minimal named-placeholder `interpolate()` helper so each locale's pattern string owns its own layout, and migrates those two spots as the first applications:

- **`interpolate()`** (`src/i18n/index.ts`): named `{x}` substitution only (no plural/select for now); unknown placeholders render as empty string.
- **Relative time**: `format.ago` → `format.relativeTime` (`"{value} ago"` / `"{value} 前"`), dropping the `getLanguage() === 'zh'` branch. This also fixes an inconsistency: the same language rendered `5m前` or `5m 前` depending on whether the stored tag was `zh` or `zh-Hans` (the latter is what the configure command saves); it is now consistently `5m 前`. `en` unchanged.
- **Absolute reset time**: `format.at` → `format.absoluteTime` (`"at {time}"` / `"{time}"`), removing the empty-string-key workaround; the preposition and spacing now live in the locale pattern.

Because these are key renames in the shared `MessageKey` type, every locale file updates to the new keys (`en` and `zh-Hans`) — this is structural, not a wording change. `en` renders identically (`5m ago`, `at 14:30`); `zh-Hans` changes only the relative-time spacing noted above (`5m前` → `5m 前`). No other locale strings change.

## Why a helper, not `isCjkLanguage()`?

The minimal fix looks like swapping `getLanguage() === 'zh'` for `isCjkLanguage()`. But a boolean CJK branch can only toggle one binary decision (space or no space) — it cannot handle word **order**. A named placeholder lets the locale decide where the value goes, with no code branch. For example, a phrase like `resets in {time}` could become `{time} 後重置` in another language (the value moves to the front) — purely in the pattern. The same mechanism makes future locales pure data, with no render-code changes.

## Testing

- [x] `npm test`
- [x] `npm run test:coverage`

Updated `tests/session-time.test.js` (zh suffix is now `5m 前`).

## Checklist

### DIFF
--- a/src/i18n/en.ts
+++ b/src/i18n/en.ts
@@ -25,14 +25,14 @@ export const en: Messages = {
   // Format
   "format.resets": "resets",
   "format.resetsIn": "resets in",
-  "format.at": "at",
+  "format.absoluteTime": "at {time}",
   "format.in": "in",
   "format.cache": "cache",
   "format.out": "out",
   "format.tok": "tok",
   "format.tokPerSec": "tok/s",
   "format.justNow": "just now",
-  "format.ago": "ago",
+  "format.relativeTime": "{value} ago",
 
   // Init
   "init.initializing": "[claude-hud] Initializing...",
--- a/src/i18n/index.ts
+++ b/src/i18n/index.ts
@@ -45,3 +45,10 @@ export function t(key: MessageKey): string {
   const canon = getCanonicalLanguage();
   return locales[canon]?.[key] ?? locales.en[key] ?? key;
 }
+
+// Minimal named-placeholder interpolation. Layout that varies by language
+// (spacing, affix position) lives in each locale's pattern string rather than in
+// render code. Unknown placeholders render as empty string (kept lenient).
+export function interpolate(pattern: string, params: Record<string, string | number>): string {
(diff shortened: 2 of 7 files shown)


UnpxreTW:
Related: #645 (add zh-Hant / zh-TW locale) — great to see Traditional Chinese landing upstream.

Small overlap to flag: both PRs touch the same `withAgo` spacing logic in `src/render/lines/session-time.ts`, so whichever merges first, the other needs a quick rebase there.

They actually compose well. This PR moves the relative-time spacing out of `session-time.ts` into each locale's `format.relativeTime` pattern — so if it lands first, #645's `zh-Hant` locale gets correct spacing just by defining `"format.relativeTime": "{value} 前"`, and the `isCjkLanguage()` tweak to `session-time.ts` in #645 wouldn't be needed.

Happy to sequence / rebase whichever way is easiest to review.

PR #649, opened 2026-07-01
feat: add opt-in SSH targets element (display.showSsh)

IvanShan177:
## Summary

Adds an opt-in, session-scoped **`ssh`** display element that shows the SSH
destinations the current session connected to, parsed from `ssh …` Bash
commands in the transcript.

```
⚡ SSH main 44.245.72.210:22 · sub#a4ad71 10.0.0.5:2222
```

Enabled via `display.showSsh` (default **off**), matching the `showTools` /
`showMcp` convention. Hidden entirely when disabled or when no `ssh` commands
are found.

## Why

When driving remote GPU/CI boxes over SSH — often several at once, including
from subagents — it's easy to lose track of which host each agent is on. This
surfaces it at a glance in the HUD, using data claude-hud already has access to
(the local transcript), with no network access.

## Behavior

- **Session-scoped.** Reads the main transcript plus the session's subagent
  transcripts at `<transcript_dir>/<session_id>/subagents/*.jsonl` (derived from
  the stdin `transcript_path`). Nothing outside the current session is read.
- **One `host:port` per source**, most-recent first: the main agent, then up to
  3 most-recently-active subagents (`sub#<short-id>`).
- **Robust parsing** (`src/ssh-targets.ts`): handles `-p`/`-pPORT`, skips
  value-taking options (`-i`, `-o`, …) so their values aren't mistaken for the
  host, ignores `ssh-keygen`/`ssh-add`, and requires an IPv4 or dotted hostname.
  Bounded tail-reads + guarded fs access, so a missing `subagents/` dir (older
  Claude Code) degrades gracefully to main-only.
- Gathered only when the element is enabled (no cost for users who leave it off).

## Files

- `src/ssh-targets.ts` — tokenizer + session collector (new)
- `src/render/ssh-line.ts` — renderer (new)
- `src/config.ts`, `src/types.ts`, `src/render/index.ts`, `src/index.ts` — wiring
  (element type, default, validation, `elementOrder`, dispatch, opt-in gather)
- `README.md`, `CHANGELOG.md` — docs
- `dist/` intentionally not committed (built by CI, per CONTRIBUTING)

## Tests

`tests/ssh-targets.test.js` (tokenizer edge cases + collector: main/subagents,
recency cap, per-source dedup, missing files) and `tests/ssh-render.test.js`
(disabled/empty → null, main + subagent formatting). All new tests pass.

> Note: a handful of pre-existing tests fail on my Windows dev box (child-process
> path resolution, cache mtime, etc.); the set is identical with and without this
> change, so no regressions were introduced. They should pass on CI/Linux.

## Notes for the maintainer

Happy to adjust the render format, the source cap (1 main + 3 subagents), or
split subagent support into a follow-up if you'd prefer a main-transcript-only
first cut.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to Claude HUD will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- Add `display.showSsh` (opt-in) — a session-scoped `ssh` element that shows the SSH destinations the session connected to, parsed from `ssh …` Bash commands in the transcript. Includes the main agent plus up to 3 subagents (read from `<session>/subagents/*.jsonl`), one `host:port` per source, most recent first, e.g. `⚡ SSH main 44.245.72.210:22 · sub#a4ad71 10.0.0.5:2222`.
+
 ## [0.3.0] - 2026-06-19
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ Chinese HUD labels are available as an explicit opt-in. English stays the defaul
 | `pathLevels` | 1-3 | 1 | Directory levels to show in project path |
 | `maxWidth` | number \| `null` | `null` | Optional fallback width used only when terminal width detection fails completely |
 | `forceMaxWidth` | boolean | false | Always use `maxWidth` when it is set, even if terminal width detection returns a smaller value |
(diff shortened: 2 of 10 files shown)


jarrodwatts:
Closing this implementation because the enabled statusline path synchronously scans up to about 21 MiB across the main and subagent transcripts on each refresh, which can run roughly every 300 ms. The file-open flow also has symlink/TOCTOU and short-read hazards, and the parser reports attempted SSH commands as connected targets. A viable redesign should reuse cached transcript parsing, perform bounded incremental reads, enforce file containment, and define consistent compact/expanded behavior.

PR #658, opened 2026-07-12
feat: add per-provider/model usage display

Kyransps12:
## Summary
- Adds `src/model-usage.ts` and wires it into the session-line and usage renderers so the statusline can break down token usage by model/provider, not just an aggregate total.

## Test plan
- [ ] `npm test` passes
- [ ] Manual check: statusline shows per-model usage breakdown in a real session

Opened as draft while I finish verifying against upstream's latest changes.


jarrodwatts:
Closing this draft on the security gate. The implementation reads the Claude OAuth access token, calls an undocumented usage endpoint from the default usage path, buffers the response without a size bound, and trusts cache fields that later reach terminal rendering. The branch also includes generated dist artifacts, an invalid 0.3.0.1 version, and no focused tests for the credential, network, or cache paths. A future implementation should wait for native stdin model-scoped data or use an explicit opt-in design with bounded I/O and hardened cache handling.

PR #667, opened 2026-07-15
feat: opt-in model-scoped weekly usage windows (e.g. Fable)

aisteri:
## Problem

Claude Code's statusline stdin only carries the generic `five_hour` / `seven_day` rate limits. Weekly quotas scoped to a specific model — e.g. the **Fable** window that the `/usage` screen shows as a separate limit — never appear in the HUD, so users on Claude 5 plans can't see how much of their premium-model quota is left.

The data exists on the OAuth usage API (the same source `/usage` uses): `limits[]` entries with `kind: "weekly_scoped"` and `scope.model.display_name`.

## Fix

Add an **opt-in** `display.showScopedUsage` (default **false**) that appends model-scoped weekly windows to the usage line in the existing window style:

```
Usage ███░░░░░░░ 33% (resets in 1h 24m) | Fable ████░░░░░░ 38% (resets in 5d 19h)
```

- Scoped windows use a distinct normal-range color (Claude orange, 208) so they're visually separate from the shared 5h/7d windows; warning/critical thresholds keep the shared palette, and `colors.usage`-style overrides still apply.
- Works with `usageCompact`, `usageValue: "remaining"`, bar/text styles, and reset-time formats — it reuses the existing `formatUsageWindowPart`/`formatCompactWindowPart` paths.
- Generic by design: any `weekly_scoped` model window the API returns is rendered, not just Fable.

## Render path never blocks

`src/scoped-usage.ts` implements a stale-while-revalidate file cache (5m TTL, 1h max age, `0700`/`0600` modes):

- Fresh cache → file read only.
- Stale/missing → serve stale data for this tick and spawn a **detached child** (`--refresh`) that reads the OAuth token from the macOS keychain (`Claude Code-credentials`) or `~/.claude/.credentials.json`, calls the usage API, and rewrites the cache. A marker file dedupes concurrent refreshes.
- The token stays inside the refresh process — never logged, rendered, or written.

Because the refresh touches the credential store, the feature is **off by default** and the README documents that requirement.

## Tests

- `tests/scoped-usage.test.js` (new): cache freshness/TTL/max-age, malformed-cache tolerance, percent clamping, `parseScopedLimits` extraction from the API shape, and `refreshScopedUsage` write/no-token/API-failure paths (fetch + token injected).
- `tests/usage-coverage.test.js`: scoped windows append after shared windows, compact mode, `usageValue: "remaining"`, and no-op when absent.
- `npm test`: 894 pass / 0 fail. Only `src/`, `tests/`, `README.md` touched — no `dist/` changes per CONTRIBUTING.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,7 @@ Simplified and Traditional Chinese HUD labels are available as explicit opt-ins.
 | `display.showDuration` | boolean | false | Show session duration `⏱️ 5m` |
 | `display.showSpeed` | boolean | false | Show output token speed `out: 42.1 tok/s` |
 | `display.showUsage` | boolean | true | Show Claude subscriber usage limits when available |
+| `display.showScopedUsage` | boolean | false | Opt-in: also show model-scoped weekly quota windows (e.g. Fable) fetched from the OAuth usage API via a background-refreshed cache. Requires access to the Claude Code credential store (macOS keychain / `~/.claude/.credentials.json`) |
 | `display.usageValue` | `percent` \| `remaining` | `percent` | Usage display format (`25%` used, or `75%` remaining) |
 | `display.usageBarEnabled` | boolean | true | Display usage as visual bar instead of text |
 | `display.usageCompact` | boolean | false | Display usage in a shorter text form such as `5h: 25% (1h 30m)`; takes precedence over `display.usageBarEnabled` |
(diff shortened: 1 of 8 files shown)


aisteri:
Re-submitting from the correct account — replaced by a new PR.

PR #681, opened 2026-07-20
docs: update instructions for current Claude Code behavior

jarrodwatts:
## Summary

Claude Code has changed quite a bit since the install/setup instructions were written. This PR updates them to match current behavior, verified against the official docs ([statusline](https://code.claude.com/docs/en/statusline.md), [plugins reference](https://code.claude.com/docs/en/plugins-reference.md)) rather than memory.

## Changes

**No more restart-after-setup** (`README.md`, `README.zh.md`, `CLAUDE.README.md`, `commands/setup.md`)
- Per current docs: *"Settings reload automatically, but changes won't appear until your next interaction with Claude Code."* The HUD appears after the next message — no restart needed, and no documented Windows difference.
- Restart is demoted to a fallback for older Claude Code versions everywhere it was previously required: the post-setup message, the Windows note, the troubleshooting sections, and the restore-previous-statusline steps.
- Setup's Step 5 no longer claims "the HUD cannot appear in the same session where setup was run" — setup now flows config → verify in one session.

**Native terminal width** (`commands/setup.md`)
- Claude Code v2.1.153+ sets `COLUMNS`/`LINES` natively before invoking the statusLine command. The generated-command rationale now documents this; the `stty size </dev/tty` probe is kept as a fallback for older versions (support floor is still v1.0.80+).

**Linux EXDEV bug is fixed upstream** (`README.md`, `README.zh.md`, `commands/setup.md`)
- [anthropics/claude-code#14799](https://github.com/anthropics/claude-code/issues/14799) is closed. The warning is demoted from "Linux users click here first" to error-triggered guidance that recommends updating Claude Code, with the `TMPDIR` workaround retained for older versions.

**Install-flow touch-ups**
- Ghost-install cleanup now points at `/reload-plugins` instead of a restart.
- Added a collapsible "Prefer the terminal?" section documenting the `claude plugin marketplace add` / `claude plugin install` CLI flow.
- `CLAUDE.README.md` was missing the `/reload-plugins` step entirely; added it.

## Notes

- Generated statusline commands are untouched — the `COLUMNS` fallback chain still matters for pre-2.1.153 versions.
- `CLAUDE_CONFIG_DIR` handling left as-is (couldn't confirm current docs mention it, but no evidence it stopped working).
- Follow-up PR planned: optionally offering `refreshInterval` during setup so time-based HUD data ticks without user interaction.

### DIFF
--- a/CLAUDE.README.md
+++ b/CLAUDE.README.md
@@ -16,12 +16,17 @@ Real-time statusline showing context usage, active tools, running agents, and to
 /plugin install claude-hud
 ```
 
+Then reload plugins (no restart needed):
+```
+/reload-plugins
+```
+
 **Step 3: Configure the statusline**
 ```
 /claude-hud:setup
 ```
 
-Done. Restart Claude Code to load the new statusLine config, then the HUD will appear.
+Done. Claude Code reloads settings automatically — the HUD appears after your next message, no restart needed. (On older Claude Code versions, restart if it doesn't show.)
 
 ---
 
@@ -67,7 +72,7 @@ This adds the statusLine configuration to the user's settings.json.
 
 Tell the user:
 - Installation complete
-- Restart Claude Code to load the statusline config, then the HUD will appear
+- Settings reload automatically — the HUD appears after their next message, no restart needed (on older Claude Code versions, restart if it doesn't show)
 - It shows: context usage, active tools, running agents, and todo progress
 
 ### Step 6: Ask About Starring (Optional)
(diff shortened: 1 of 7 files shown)


jarrodwatts:
Added 201f0d6: also corrected the "updates every ~300ms" claim in README.md, README.zh.md, CLAUDE.md, and CLAUDE.README.md — per current docs the statusline runs after each interaction with a 300ms debounce, not on a fixed polling loop.

jarrodwatts:
Subagent review completed — no blockers, all factual claims verified against the live docs (auto-reload wording, COLUMNS/LINES v2.1.153 requirement, /reload-plugins hot-reload, EXDEV #14799 closed 2026-02-13, 300ms debounce). Addressed the findings in adac375:
- Fixed stale `~300ms polling loop` comments in `src/index.ts` / `src/context-cache.ts` that contradicted the corrected docs
- Added the missing `/reload-plugins` step to CLAUDE.README.md's LLM-agent workflow
- Added vim-mode toggles to the interaction-trigger lists; removed a double blank line

Deliberately skipped: modernizing the EXDEV workaround to `claude plugin install` — machines still hitting EXDEV are on old Claude Code versions that may predate that CLI, so the slash-command form is the safer instruction.

PR #682, opened 2026-07-20
feat: offer statusLine refreshInterval during setup

jarrodwatts:
## Summary

Claude Code supports an optional [`refreshInterval`](https://code.claude.com/docs/en/statusline.md) key (seconds, minimum 1) on the `statusLine` settings object that re-runs the statusline command on a timer. Without it, the HUD only re-renders after an interaction — so time-based data (session duration, usage reset countdowns, the prompt-cache countdown) goes stale between messages.

Follow-up to #681.

## Changes

- **`commands/setup.md`**: new Step 4.5 asks the user whether to enable auto-refresh (every 5s recommended / every 1s / no timer) and merges `refreshInterval` into the existing `statusLine` object — preserving `type`/`command`, reusing Step 3's JSON-safety and no-BOM rules, and removing the key if the user explicitly opts out on a re-run.
- **`README.md` / `README.zh.md`**: new "Auto-Refresh" section under Configuration with an example snippet.

## Notes

- 5 seconds is the recommended default since each refresh re-runs the full HUD command (runtime startup, transcript parse, git status); 1 second is offered for visibly smooth countdowns.
- A `hideVimModeIndicator` follow-up was considered but skipped: the HUD doesn't render `vim.mode`, so there's nothing to hide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,7 +369,7 @@ Example fallback snapshot:
 
 ### Auto-Refresh
 
-Claude Code only re-runs the statusline after an interaction (a new message, `/compact` finishing, a permission-mode change), so time-based HUD info — session duration, usage reset countdowns, the prompt-cache countdown — goes stale between messages. To keep it ticking, add `refreshInterval` (seconds, minimum 1) to the `statusLine` entry in `~/.claude/settings.json`:
+Claude Code only re-runs the statusline after an interaction (a new assistant message, `/compact` finishing, a permission-mode change, or a vim-mode toggle), so time-based HUD info — session duration, usage reset countdowns, the prompt-cache countdown — goes stale between messages. To keep it ticking, add `refreshInterval` (seconds, minimum 1) to the `statusLine` entry in `~/.claude/settings.json`:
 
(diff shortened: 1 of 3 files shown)


jarrodwatts:
Subagent review completed — no blockers. Verified `refreshInterval` semantics against the statusline docs, and confirmed in `src/` that the HUD renders genuinely time-based data (usage reset countdown default-on; session duration and prompt-cache countdown opt-in) with no internal caching that would defeat a refresh. Addressed the findings in 8e8fa4f:
- "Other" with a numeric interval is now honored (clamped to min 1) instead of ignored
- Clarified that refreshInterval hot-applies via automatic settings reload (restart as fallback) so Step 5 doesn't misread a not-yet-ticking timer as a failure
- Noted which time-based displays are default-on vs opt-in
- Completed the interaction-trigger lists (assistant messages, vim-mode toggles)

Skipped: rewording Step 3's "complete Step 4" restart message — #681 rewrites that line, so touching it here would just manufacture a merge conflict.

PR #702, opened 2026-08-06
prompt cache fixes/display  - don't count subagents, detect the TTL, show expiry time

bnorquist:
## Summary

Three fixes to the opt-in `display.showPromptCache` element. Each is a separate commit.

Today, with `showPromptCache` enabled, the element can report a warm cache that has already expired, expire 55 minutes early, or freeze on a stale number — depending on which of the three you hit.

## 1. Subagent responses refresh the cache clock (`095f41e`)

`lastAssistantResponseAt` is assigned on any record with `type === 'assistant'`. Sidechain (subagent) records are interleaved into the main session's transcript, but they run against their own prompt cache and do not refresh the main session's.

One `Task` call is enough: the subagent's response pushes the deadline forward by however long the subagent ran, and the HUD then shows time the session does not have. Fixed by excluding `isSidechain` records from that one assignment — session token totals and every other aggregate still count them.

## 2. The TTL is configured rather than detected (`d0c663d`)

`display.promptCacheTtlSeconds` defaults to `300`. Anyone whose session actually runs on the 1-hour tier and hasn't changed that setting sees the cache read `expired` 55 minutes before it is — in the direction that makes them rush a request they didn't need to send.

The tier doesn't have to be guessed. Every cache write records which one it used:

```json
"cache_creation": { "ephemeral_1h_input_tokens": 1128, "ephemeral_5m_input_tokens": 0 }
```

`parseTranscript` now reads that from each main-chain assistant response into `TranscriptData.promptCacheTtlSeconds`, and the renderer prefers it. Because it's re-read per response, a session that changes tier mid-run (e.g. dropping to 5m on usage overage) is followed rather than pinned.

- Mixed tiers are representable across multiple cache breakpoints; the shortest wins, being the first part of the prefix to lapse.
- `TRANSCRIPT_CACHE_VERSION` 13 → 14, since v13 snapshots carry no TTL field. One-time cache invalidation on upgrade.

Semantic change: `display.promptCacheTtlSeconds` becomes a fallback/override rather than *the* TTL. Detection wins because it's read from the request that actually wrote the cache, so it's a fact about the session rather than a guess. 

## 3. A countdown can't stay correct here (`fbef48b`)

The statusline only re-renders while Claude Code is active. Sitting idle at the prompt — exactly when the cache is draining — nothing triggers a repaint, so the countdown freezes at whatever it last displayed and keeps showing it long after the cache is gone.

I measured it on 0.1.0 by logging every statusline invocation: **4 calls inside a 26-second burst of activity, then zero across the next 2 minutes of idle.**

A relative duration is only meaningful if something redraws it; an absolute time is still true however stale the render is. So the value becomes the expiry clock time, reusing the existing `formatAbsoluteTime` and its locale pattern so it matches how usage reset times already render:

```
Cache ⏱ at 14:30      →  still true five minutes later, and an hour later
Cache ⏱ expired       →  once elapsed
```

## Verification

Against a real transcript on the 1h tier, rendering at three points after the last response:

| Rendered | This PR | Before |
|---|---|---|
| 1 min after | `Cache ⏱ at 05:05 PM` | `Cache ⏱ at 04:10 PM` (55 min early) |
| 59 min after | `Cache ⏱ at 05:05 PM` | `Cache ⏱ expired` (still warm) |
| 2 hrs after | `Cache ⏱ expired` | `Cache ⏱ expired` |

The middle row is the point of #3: the value doesn't change between renders, so there's nothing to go stale.

Tier detection was checked against 11,530 assistant records across 88 local transcripts — 11,518 single-tier, 12 pure cache reads (the carry-forward path), 0 mixed.

## Testing

`npm test`: **976 passing, 0 failing** (was 969 before; 5 pre-existing skips unchanged).

New coverage in `tests/core.test.js` (sidechain exclusion, tier detection, carry-forward, mid-session change, subagent tier not adopted, no-cache-write case) and `tests/prompt-cache.test.js` (TTL precedence, expiry rendering, expired state). `tests/render.test.js` assertions updated for the new value format. Clock assertions derive the expected string via `toLocaleTimeString` rather than hardcoding, following `tests/format-reset-time.test.js`, so they're timezone-independent.

`src/`, `tests/` and `README.md` only — no `dist/`, per CONTRIBUTING.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ After choosing a preset, you can turn individual elements on or off.
 ### Manual Configuration
 
 Edit `~/.claude/plugins/claude-hud/config.json` directly for advanced settings such as `colors.*`,
-`pathLevels`, `maxWidth`, threshold overrides, `display.timeFormat`, and `display.hourCycle`. Running `/claude-hud:configure`
+`pathLevels`, `maxWidth`, threshold overrides, `display.timeFormat`, `display.hourCycle`, and `display.promptCacheTtlSeconds`. Running `/claude-hud:configure`
 preserves those manual settings while still letting you change `language`, layout, and the common
 guided toggles.
 
@@ -232,6 +232,7 @@ Simplified and Traditional Chinese HUD labels are available as explicit opt-ins.
 | `display.showClaudeCodeVersion` | boolean | false | Show the installed Claude Code version, e.g. `CC v2.1.81` |
 | `display.showMemoryUsage` | boolean | false | Show an approximate system RAM usage line in expanded layout |
 | `display.showPromptCache` | boolean | false | Show the wall-clock time the session's prompt cache expires, read from the transcript |
+| `display.promptCacheTtlSeconds` | number | `300` | Compatibility fallback used only when the transcript has not reported a 5-minute or 1-hour cache tier |
 | `colors.context` | color value | `green` | Base color for the context bar and context percentage |
 | `colors.usage` | color value | `brightBlue` | Base color for usage bars and percentages below warning thresholds |
 | `colors.warning` | color value | `yellow` | Warning color for context thresholds and usage warning text |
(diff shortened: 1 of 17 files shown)


jarrodwatts:
Maintainer review: the main-chain cache clock, tier detection, and absolute expiry ideas are useful, and the review found no new network, shell, dependency, generated-artifact, credential, or persistence behavior. The current draft is not mergeable yet because it conflicts with main and needs a careful current-main rewrite. Required integration changes: use transcript cache version 16 because main is already 15 and 14 was previously used for another schema; accept cached TTL values only when exactly 300 or 3600; keep a prompt-cache-specific main-chain timestamp so showLastResponseAt semantics do not change; preserve the current WallClockOptions, including hour cycle and seconds; resolve the README versus code precedence for detected TTL and promptCacheTtlSeconds; and account for the fact that cache lifetime begins with the request that reads or writes the cache, not the later assistant response. Please keep dist excluded. Once rebased and updated, this can receive a fresh security and regression gate.

bnorquist:
Thanks for the review addressed all of them. One decision I made of note: 

On the promptCacheTtlSeconds precedence question, we removed the setting instead of resolving it either way. The element stays hidden until the session has a main-chain response, and that same response is what reports the tier — across 13,309 local main-chain records, 13,292 carry the per-tier cache_creation breakdown and not one cache write lacked it. So once detection exists, the setting is read in a window that is close to unreachable: as a fallback it would almost never apply, and as an override it could only make the HUD disagree with the request that actually wrote the cache. A documented number that cannot change what you see is worse than no number, so the 5-minute tier became a constant and the config surface went away.

Happy to discuss that further if you want to handle it a different way.

gr4cza:
@jarrodwatts, in #682, you added an option to set `refreshInterval`, which worked flawlessly to update the cache’s remaining time. Now there is no option to have a countdown. Can you add back a setting to choose between a countdown and an exact end time?

PR #705, opened 2026-08-11
fix(version): resolve Windows cmd.exe from SystemRoot

jhuang-tw:
## Summary

Fix Windows Claude Code version detection when Windows is installed on a drive other than `C:`.

`src/version.ts` previously hard-coded:

`C:\Windows\System32\cmd.exe`

for launching `.cmd` / `.bat` Claude binaries. On systems where Windows is installed under another drive, such as `D:\Windows`, the version probe can fail even though Claude Code itself is installed and working.

This change:

- resolves `cmd.exe` from `SystemRoot` / `SYSTEMROOT`
- accepts only drive-absolute Windows paths
- rejects malformed, relative, or null-byte-containing values
- preserves `C:\Windows\System32\cmd.exe` as the safe fallback
- does not trust `COMSPEC` or PATH for the shell location

Example:

```text
SystemRoot=D:\Windows
→ D:\Windows\System32\cmd.exe
```

The existing behavior for normal `C:\Windows` installations is unchanged.

## Testing

- [x] Added regression coverage for Windows installed under `D:\Windows`
- [x] Added coverage ensuring relative `SystemRoot` values fall back safely
- [x] Verified the branch contains exactly one commit relative to upstream `main`
- [ ] `npm test` — CI
- [ ] `npm run test:coverage` — CI

## Checklist

- [x] Tests updated
- [x] No documentation change needed; this is an internal platform compatibility fix
- [x] No `dist/` changes

### DIFF
--- a/src/version.ts
+++ b/src/version.ts
@@ -40,12 +40,13 @@ type ClaudeVersionInvocation = {
 };
 
 const CACHE_FILENAME = '.claude-code-version-cache.json';
+const DEFAULT_WINDOWS_CMD = 'C:\\Windows\\System32\\cmd.exe';
 const defaultExecFile: ExecFileImpl = promisify(execFile) as ExecFileImpl;
 
 let execFileImpl: ExecFileImpl = defaultExecFile;
 let resolveClaudeBinaryImpl: () => ClaudeBinaryInfo | null = resolveClaudeBinaryFromPath;
 let platformImpl: () => NodeJS.Platform = () => process.platform;
-let windowsCmdImpl: () => string = () => 'C:\\Windows\\System32\\cmd.exe';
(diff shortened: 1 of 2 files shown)


jarrodwatts:
Thanks for the focused report and patch. I cannot merge this implementation because SystemRoot is inherited process state, and accepting any drive-absolute value makes the interpreter path attacker-selectable. A value pointing at an arbitrary writable tree or containing parent traversal can select a planted System32\\cmd.exe for execution. That reverses the executable-path trust hardening from #487. The current tests cover D:\\Windows and a relative fallback, but not arbitrary absolute roots or traversal. We will retain the fixed trusted interpreter path unless a Windows-native resolution design can establish the OS directory without trusting inherited environment state.

PR #708, opened 2026-08-12
feat: add plancost segment for third-party coding-plan usage/balance

Pgooone:
## Summary

Adds a **plancost** first-line segment that shows coding-plan usage or account balance for third-party model providers (Kimi For Coding, DeepSeek, Zhipu GLM), queried directly from each provider API.

```
🟡 Kimi 5h 15% (03:44) · week 69% (08/13)      ← coding-plan windows
💰 DS ¥40.96                                   ← account balance
```

Health markers: 🟢 < 60% / 🟡 60–84% / 🔴 ≥ 85% (worst window wins).

### Relationship to the existing usage / external snapshot features

The official `rate_limits` display requires an Anthropic subscriber login — API-key and proxy users get no `rate_limits` on stdin. The external usage snapshot path (#690) covers model-scoped windows for Anthropic models via a sidecar file that an external feeder writes. Plancost is the complementary path for **third-party providers**: it fetches the provider's own usage/balance API itself (with a 5-minute disk cache), so users of unified relays that mix e.g. Kimi K3 and DeepSeek can see both quotas without running a feeder.

### Key design points

- **Provider model matching**: `displayMode: "auto"` shows the provider whose `models` prefix matches the **actually served model** (transcript `message.model`, stdin fallback); `"all"` shows every configured provider. Transcript matching skips future sidechain (subagent) records.
- **Caching**: one JSON file per provider under `~/.claude/plugins/claude-hud/plancost-cache/`, 5-minute TTL, keyed by a hash of the apiKey — swapping keys invalidates stale data immediately; network failure falls back to stale cache; the segment silently hides when nothing is available.
- **Safety**: provider responses are sanitized before display; the module never throws; keys are plain-text in `config.json` (documented).
- **Setup wizard**: `/claude-hud:setup` Step 4 and `/claude-hud:configure` Q7 gain a bilingual (EN / 简体中文) wizard for enabling providers, entering keys, display mode, and `projectLineOrder` placement. The wizard detects an **Anthropic official OAuth login** (no custom base URL + `oauthAccount` present) and asks whether third-party plancost is still wanted, since the official `rate_limits` usage is already shown.

## Testing

- [x] `npm test`
- [x] `npm run test:coverage`
- [x] Tests updated (35 new cases: provider parsing, model matching, cache/TTL/key-hash behavior, stale fallback, rendering format, i18n)

Docs updated: README (Plancost section + Security Notes), CHANGELOG ([Unreleased]).

Note: `dist/` is intentionally excluded per CONTRIBUTING — CI compiles and commits it after merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@ All notable changes to Claude HUD will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- `plancost` first-line segment showing third-party coding-plan usage or
+  account balance (Kimi / DeepSeek / Zhipu GLM), fetched directly from the
+  provider APIs with a 5-minute per-provider disk cache and stale-cache
+  fallback. Configured via the top-level `plancost` block; `displayMode`
+  `auto` matches the provider to the actually served model, `all` shows every
+  configured provider (#PR).
+- `/claude-hud:setup` and `/claude-hud:configure` wizard questions now support
+  Simplified Chinese alongside English, driven by the HUD `language` config.
+  The setup wizard detects an Anthropic official OAuth login and asks whether
+  third-party plancost should still be enabled (#PR).
+
 ## [0.7.1] - 2026-08-11
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -308,6 +308,89 @@ Set `display.showResetLabel` to `false` if you want shorter usage countdowns suc
 
 Set `display.usageCompact` to `true` if you want the shorter usage-only form, for example `5h: 25% (1h 30m)`. Compact usage takes precedence over `display.usageBarEnabled`.
 
+### Plancost (third-party model providers)
+
+A `plancost` first-line segment shows **coding-plan usage or account balance
+for third-party domestic model providers**, queried directly from the provider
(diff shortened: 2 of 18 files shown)


wuchenkang:
@Pgooone Any plan of adding support for volcengine coding plan (https://www.volcengine.com/activity/codingplan)?

Pgooone:
> @Pgooone Any plan of adding support for volcengine coding plan (https://www.volcengine.com/activity/codingplan)?

Yes, my fork project has added volcanic support.https://github.com/Pgooone/claude-hud

jarrodwatts:
Thanks for the substantial work and test effort. I am closing this version because it changes Claude HUD's trust boundary in ways that are not safe to ship. The setup flow asks users to paste spend-capable API keys into the model conversation, the GLM endpoint override can send its Authorization header to an arbitrary host, and the disk cache follows symlinks, reads unbounded unvalidated data, uses a predictable temp path, and can render poisoned cache strings without a final terminal-safety check. The README also retains the current local-only/no-network security guarantee while this change adds live network requests. Fixing these issues requires a maintainer-owned redesign, not a small patch to this branch. A future proposal should use the existing external snapshot boundary or a local non-LLM secret entry path, pin credential destinations, and reuse the repository's hardened cache contract. Volcengine support should remain separate from that security design.

PR #709, opened 2026-08-13
feat: add MiniMax and Zhipu API provider quota display

ongyishen:
## Summary
    This PR adds support for displaying real-time quota and usage information for **MiniMax** and **Zhipu** APIs directly in the HUD.

    - Added a new `providerQuota` configuration setting that accepts API keys for both providers.
    - Integrated `fetchProviderQuotas` using the official endpoints:
      - MiniMax: `coding_plan/remains`
      - Zhipu: `/api/monitor/usage/quota/limit`
    - Safely extracts the remaining percentage from both API responses (since absolute token counts often return as 0 or undefined).
    - Renders the quota as a compact table line (e.g., `MiniMax: 5h:84% wk:100% │ Zhipu: 5h:77%`) using the existing line-rendering architecture.

 ## Testing

    - [x] `npm test`
    - [x] `npm run test:coverage`

    *(Note: The test suites have been verified against the new changes. No existing tests were broken because this relies strictly on an optional config property.)*

  ## Checklist


jarrodwatts:
Thanks for the contribution. I am closing this as unsafe, under-tested overlap with #708. Remote MiniMax error text reaches the terminal without sanitization, the same Zhipu credential is sent across three fallback domains, and the shared cache follows symlinks, accepts unbounded unvalidated data, uses default permissions, and is not keyed to the configured account. The committed dist is also broken: dist/index.js imports provider-quota.js and the lines barrel exports it, but those generated modules are absent, so the published entrypoint would fail before the opt-in config can help. There are no source feature tests or docs in the PR despite the body claiming tests. This needs a maintainer-owned secure design around the existing external snapshot path rather than merging this branch.

PR #720, opened 2026-08-20
feat(usage): read model-scoped windows from the Claude CLI usage cache, opt-in auto-refresh

sebyul2:
## Problem

`/usage` in Claude Code shows a weekly window scoped to a specific model (the "Fable" weekly quota, for example) alongside the account-wide 5h and 7d windows. #669 taught the HUD to render those scoped windows, reading them from `rate_limits.model_scoped` on statusline stdin.

Claude Code does not send that field over statusline stdin. So the rendering added in #669 has no data source out of the box, and nothing shows.

The README's only alternative is `display.externalUsagePath`: point the HUD at a JSON snapshot you keep fresh yourself with a cron/launchd job piping `get_usage` through `jq`. It works, but it is a scheduled job most people will never set up.

Two things changed on the Claude Code side in recent releases, both verified empirically against the current CLI binary:

1. The CLI persists every `/usage` fetch into `{CLAUDE_CONFIG_DIR}.json` under `cachedUsageUtilization`, including the model-scoped weekly windows, each carrying `scope.model.display_name`.
2. Headless `claude -p /usage` finishes in a few seconds, spends zero tokens, invokes no model, and rewrites that same cache.

So the data the HUD wants is already sitting in a file the CLI maintains for itself, and there is a zero-cost way to make the CLI refresh it.

## What this adds

Two separate things, deliberately split:

**Reading the cache is unconditional.** The HUD now always consults `cachedUsageUtilization` as the last scoped-window fallback. No configuration: if you have run `/usage` recently, the scoped windows appear. This is a plain local file read of a file the HUD already opens — `{CLAUDE_CONFIG_DIR}.json` is the same file it reads for auth info today — so it adds no new capability, only a new field read from a path already in use.

**Auto-refresh is opt-in**, via `display.refreshModelScopedUsage` (default `false`):

```json
{ "display": { "refreshModelScopedUsage": true } }
```

Enabling it does two things: the HUD keeps the cache fresh by spawning a detached headless `claude -p /usage` when the cache is older than five minutes, and — because that freshness is then guaranteed — the CLI cache is promoted above external snapshots as the more trustworthy source.

Priority for scoped windows:

| | order |
|---|---|
| refresh **off** (default) | stdin → external snapshot → CLI cache |
| refresh **on** | stdin → CLI cache → external snapshot |

Stdin always wins. One subtlety worth calling out: an explicit *empty* `model_scoped` array on stdin is preserved as-is, because that is Claude Code stating "there are no scoped windows". Only a genuinely missing field falls through to the fallbacks. That keeps the retirement story exact — the day Claude Code forwards the field, this whole block stops contributing, whether the array has entries or not.

## How it works

1. On render, the HUD reads `cachedUsageUtilization` from `{CLAUDE_CONFIG_DIR}.json` and projects its model-scoped `limits[]` entries into scoped windows. No subprocess in this path.
2. Fresh entries render, through the same `parseScopedWindows` bounds the stdin path uses, so label sanitization, entry caps, percent clamping, and reset-string validation all still apply. Anything older than one hour is not rendered at all — that is the staleness bound the CLI itself applies to this cache, so the HUD never shows data `/usage` would no longer trust.
3. **Only if `refreshModelScopedUsage` is enabled** and the cache is older than five minutes, the HUD spawns one detached, headless `claude -p /usage`. Five minutes is the CLI's own cache write-throttle interval, so polling faster cannot produce newer data.
4. The CLI does its own fetch and rewrites its own cache file. The next render picks it up.

Model names come from the cache verbatim; there is no hardcoded model list, so any future model-scoped window works without a code change.

## Security

The README promises the HUD is local-only and never fetches anything. Both halves of this keep that promise, and the always-on half is the passive one:

- **The always-on part is a local file read and nothing else.** No subprocess, no network, no new file path — it reads one more field out of a JSON file the HUD already opens for auth info.
- **The subprocess is strictly opt-in.** With the default config, nothing is ever spawned; the HUD only reads whatever the CLI happens to have cached.
- **The HUD never reads or handles credentials.** When refresh is enabled, the CLI performs its own fetch with its own auth, exactly as if you had typed `/usage` yourself. The cache file is the only channel between the two.
- **No tokens are spent and no model is invoked.** `claude -p /usage` runs the local command and exits.
- **The subprocess is spawned without a shell**, detached, with `stdio: 'ignore'`. Nothing is parsed from its output; spawn failures are swallowed and never break a render.
- **No stampede.** Statusline renders happen far more often than a headless `/usage` completes, and several sessions share one binary. A private marker file under `~/.claude/plugins/claude-hud` (dir `0700`, file `0600`) is claimed *before* spawning, so concurrent renders cannot fan out into duplicate subprocesses.

The exit condition is unchanged: once Claude Code starts forwarding `rate_limits.model_scoped` on stdin, stdin wins and this path stops executing. It can then be deleted without touching the rendering from #669.

## Tests

`tests/cli-usage.test.js`, 24 tests:

- cache parsing and projection (multiple models, entries missing a display name, null percent, invalid reset strings, missing file / missing cache block / bad timestamp)
- freshness windows (fresh, stale-but-usable, past the one-hour bound, timestamps from the future)
- refresh scheduling (stale schedules exactly one, fresh schedules none, missing cache primes the first render, and reading still works with refresh disabled without spawning anything)
- stampede suppression via the marker (recent marker suppresses, expired marker retries, consecutive stale renders spawn once)
- spawn failures swallowed
- config validation of the opt-in
- `main()` integration: backfill onto existing stdin usage, build usage from the cache alone, stdin scoped windows win, the refresh flag is forwarded through, and priority in all three directions — external snapshot wins when refresh is off, CLI cache wins when refresh is on, snapshot is kept when refresh is on but the cache is empty

One honest note on test isolation: five pre-existing `main()` usage tests in `tests/index.test.js` now stub `getScopedUsageFromCliCache` to return `null`. Because the read is unconditional, those tests would otherwise pick up a real populated `claude.json` on a developer's machine and stop being deterministic. The stub is the whole change to that file.

Full suite: 1103 tests, 1097 pass, 0 fail, 6 pre-existing skips.

Per CONTRIBUTING, only `src/`, `tests/`, and `README.md` are touched — no `dist/` changes. README gets one options-table row and two paragraphs in the usage section next to the existing `externalUsagePath` guidance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,7 @@ Simplified and Traditional Chinese HUD labels are available as explicit opt-ins.
 | `display.externalUsagePath` | string | `""` | Optional absolute path to a local usage snapshot file. Relative paths are ignored. When stdin `rate_limits` are present, `balance_label` is appended and `model_scoped` windows fill in when stdin lacks them; when stdin windows are missing, valid usage windows can be used as a fallback |
 | `display.externalUsageWritePath` | string | `""` | Optional absolute `.json` path in an existing directory. When stdin `rate_limits` exists, ClaudeHUD writes a private snapshot for other local tools. Relative paths, non-json files, and missing parent directories are ignored |
 | `display.externalUsageFreshnessMs` | number | `300000` | Maximum allowed age for the external usage snapshot before it is ignored |
+| `display.refreshModelScopedUsage` | boolean | false | The HUD always reads model-scoped weekly windows (e.g. Fable) from the usage cache the Claude CLI keeps in `{CLAUDE_CONFIG_DIR}.json` as the last fallback. This opt-in keeps that cache fresh via a detached headless `claude -p /usage` when it is older than five minutes, which promotes the cache above external snapshots. Stdin always wins |
 | `display.showTokenBreakdown` | boolean | true | Show token details at high context (85%+) |
 | `display.showTools` | boolean | false | Show tools activity line |
 | `display.showSkills` | boolean | false | Show active Skills detected from `Skill` tool invocations |
@@ -301,6 +302,10 @@ The snapshot may also carry `model_scoped` windows using the same shape Claude C
 ```
 
 One zero-credential way to produce such a snapshot is Claude Code's own `get_usage` control request, which returns `rate_limits.model_scoped` without spending tokens; a scheduled job can pipe it through `jq` into the snapshot file. The HUD itself never fetches anything: it only reads the file.
+
+ClaudeHUD also reads the usage cache the Claude CLI already maintains in `{CLAUDE_CONFIG_DIR}.json` (written on every `/usage` fetch, including model-scoped windows). This needs no setup: if you have run `/usage` recently, its scoped windows appear as the last fallback, behind stdin and external snapshots, and entries older than the one hour the CLI itself trusts are never rendered.
+
+If you would rather not maintain a scheduled job, `display.refreshModelScopedUsage` keeps that cache fresh automatically: when it is older than five minutes — the interval the CLI itself throttles cache writes to — ClaudeHUD spawns one detached, headless `claude -p /usage` so the CLI refreshes its own cache, and that freshness guarantee promotes the cache above external snapshots (stdin still always wins). The subprocess spends no tokens and no model is invoked; the CLI resolves authentication itself and ClaudeHUD never sees credentials or talks to the network — the cache file is the only channel. A private marker under `~/.claude/plugins/claude-hud` prevents concurrent renders from stampeding duplicate refreshes, and because stdin scoped windows take priority outright, the whole path retires naturally once Claude Code forwards `rate_limits.model_scoped` on stdin.
 
 Set `display.externalUsageWritePath` if you want ClaudeHUD to write the official stdin `rate_limits` into a local snapshot for other tools. The path must be absolute, end in `.json`, and live in an existing directory. ClaudeHUD writes the file with private permissions and ignores invalid paths quietly.
 
--- a/src/cli-usage.ts
+++ b/src/cli-usage.ts
@@ -1,0 +1,231 @@
+import { spawn } from 'node:child_process';
+import * as fs from 'node:fs';
+import * as os from 'node:os';
+import * as path from 'node:path';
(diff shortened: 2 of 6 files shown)


kuk1song:
I wrote the external-snapshot path (#690) that this PR re-orders priority around, so I went and checked the premise on my own machine.

**The cache premise holds.** On Claude Code 2.1.220, `~/.claude.json → cachedUsageUtilization` does carry the model-scoped weekly window in exactly the shape `readCliUsageCache` expects: `scope.model.display_name: "Fable"`, a `percent`, and an ISO `resets_at`. Re-checked today, still there.

**One caveat on freshness, which I can't settle from my own machine.** I run a scheduled `get_usage` feeder for the external-snapshot path, and it turns out that refreshes `cachedUsageUtilization` as well: my cache's `fetchedAtMs` lands within seconds of every poll. So I have no clean read on how fresh that cache is for a normal setup, and I wouldn't want my box quoted as evidence that the unconditional fallback lights up for most people. Given the one-hour bound, it might be worth saying in the README what to expect when you haven't opened `/usage` in a while.

**The spawn is as cheap as advertised.** I ran `claude -p /usage` here: `total_cost_usd: 0`, `num_turns: 0`, all four token counters at zero, about 2.4s, and it did rewrite `cachedUsageUtilization` right after. One thing that bit me in my own feeder though: `spawn('claude', ...)` goes through PATH, and a Claude Code started from a GUI or an IDE can have a PATH without `~/.local/bin`. It fails safe, the error handler just logs and the unconditional cache read still works, but the opt-in can quietly do nothing on those setups.

**The priority chain reads right.** stdin always wins, and the cache only outranks the external snapshot when `refreshModelScopedUsage` guarantees its freshness. That keeps #690's semantics intact, and both paths retire cleanly if anthropics/claude-code#73770 ever forwards `model_scoped` on stdin.

**One gap this surfaces.** Once scoped windows show up by default, there's no way to hide only those: `display.showUsage` takes the 5h/7d windows with it, and `sevenDayThreshold` only gates the 7d bar. I opened #728 for that. It defaults to `true` so it changes nothing on its own, and it gates every source, including this PR's cache path.

Issue #576, opened 2026-05-29
Context bar drops to 0% during assistant streaming when current_usage and totals are both empty


Cedriccmh:
## Summary

When the assistant is streaming tokens, Claude Code occasionally feeds the statusline a frame where everything is zero:

```json
{
  "context_window": {
    "used_percentage": 0,
    "current_usage": null,
    "context_window_size": 1000000
  }
}
```

`isSuspiciousZero` in `src/context-cache.ts` short-circuits this case because the predicate requires `total_input_tokens > 0 || total_output_tokens > 0` (added in #508). Mid-stream Claude Code is happy to emit a frame where those totals are also `0`/`undefined`, so the cache fallback never kicks in and the HUD renders `Context ░░░░░░░░░░ 0%` for the entire generation, then snaps back to the real value when the response completes.

Reproduces against the current `main` (v0.1.0) on Windows 11, Claude Code `2.1.153`, Opus 4.7. The cache file under `plugins/claude-hud/context-cache/<sha>.json` already contains a known-good snapshot (e.g. `used_percentage: 9`) from the previous tick — we just refuse to use it.

## Minimal repro

```bash
echo '{
  "transcript_path": "/tmp/x.jsonl",
  "model": {"id":"claude-opus-4-7","display_name":"Opus 4.7"},
  "context_window": {
    "used_percentage": 0,
    "remaining_percentage": 100,
    "current_usage": null,
    "context_window_size": 1000000
  }
}' | node dist/index.js
```

Expected: HUD shows the last-known-good percentage from the session-scoped cache.
Actual: HUD shows `0%`.

Same outcome with `current_usage: {input_tokens:0, output_tokens:0, cache_creation_input_tokens:0, cache_read_input_tokens:0}` and with `context_window` missing the usage fields entirely.

## Root cause

`isSuspiciousZero` requires accumulated totals to be non-zero before treating the frame as a glitch:

```js
const totalInputTokens = contextWindow.total_input_tokens ?? 0;
const totalOutputTokens = contextWindow.total_output_tokens ?? 0;
return totalInputTokens > 0 || totalOutputTokens > 0;
```

[...]

PR #743, opened 2026-09-01
fix(context): synthesize used_percentage before the cache write gate when current_usage is real

### LINKED ISSUE #742: Context cache silently freezes when used_percentage is 0 but current_usage is real — statusline and cache diverge (follow-up to #430)
## Summary

When Claude Code emits a frame where `context_window.used_percentage` is `0` but `context_window.current_usage` already has real, non-zero token counts, the statusline and the session-scoped context cache diverge and **stay diverged**.

This zero-with-real-usage frame is already known to the codebase: the comment on `getNativePercent` (`src/stdin.ts:159-163`) describes it for the fresh-session case, and PR #430 ("show initial context tokens when used_percentage is 0", fixing #417) added the render-side fallback precisely for it. **This report is the other half of that same fix.** #430 corrected what the statusline *displays*; the cache-write gate was not updated in the same pass and still tests the raw `used_percentage`, so it skips the write on exactly the frames #430 taught the renderer to handle. Our field observation also suggests the frame is not limited to session start — we saw the divergence with substantial usage already accumulated, well into a session (details below).

The two halves:

- The rendered statusline correctly falls back to a live percentage computed from `current_usage` and displays it.
- The context-cache write is skipped entirely for that frame, because the write gate checks the raw (untouched) `used_percentage`, not the live fallback value.

If native `used_percentage` keeps reading `0` for a while, the cache stays frozen at whatever it last held, while the rendered number keeps moving. Any other consumer that reads the cache file (e.g. a script gating context-fill behavior on it) then sees a stale, materially **lower** number that still looks fresh going by `saved_at`.

Observed in the field: statusline showing 61%, cache file showing 38%, same session, same moment — a 23-point gap that a "checked the cache 30 seconds ago" consumer would never suspect.

Found on commit `939eb66` (current `main`, 2 commits past the v0.8.0 tag). Reproduces on every version back through at least v0.6.0 — the relevant functions (`getNativePercent`, `hasGoodContext`, `writeCache`) are unchanged across that range.

## Root cause

Two files, four functions, one untouched raw value:

1. **`src/stdin.ts:165-171`** — `getNativePercent()` treats `used_percentage <= 0` as "not populated" and returns `null`:

   ```ts
   function getNativePercent(stdin: StdinData): number | null {
     const nativePercent = stdin.context_window?.used_percentage;
     if (typeof nativePercent === 'number' && !Number.isNaN(nativePercent) && nativePercent > 0) {
       return Math.min(100, Math.max(0, Math.round(nativePercent)));
     }
     return null;
   }
   ```

2. **`src/stdin.ts:180-192`** (inside `getContextPercent`, and the same pattern in `getBufferedPercent`) — on `null`, it recomputes and **displays** a live percentage from `current_usage`:

   ```ts
   const native = getNativePercent(stdin);
   if (native !== null) {
     return native;
   }
   // Fallback: manual calculation without buffer
   const size = stdin.context_window?.context_window_size;
   if (!size || size <= 0) {
     return 0;
   }
   const totalTokens = getTotalTokens(stdin);
   return Math.min(100, Math.round((totalTokens / size) * 100));
   ```

3. **`src/context-cache.ts:249-255`** — `hasGoodContext()`, which gates the cache write, checks the same **raw** `used_percentage` and never sees the fallback value computed above:

   ```ts
   function hasGoodContext(contextWindow: ContextWindow): boolean {
     return (
       (contextWindow.context_window_size ?? 0) > 0 &&
       typeof contextWindow.used_percentage === "number" &&
       contextWindow.used_percentage > 0
     );
   }
   ```

4. **`src/context-cache.ts:338-339`** — so the write is skipped on exactly the frames where the renderer computed something worth caching:

   ```ts
   if (hasGoodContext(contextWindow)) {
     writeCache(homeDir, transcriptPath, contextWindow, now, sessionName);
     ...
   }
   ```

Note this is **not** the same bug `isSuspiciousZero()` / `#508` / `#576` cover. Those handle `used_percentage: 0` combined with an **empty** `current_usage` (restoring from cache on a genuine glitch or fresh-session frame). This report is the opposite combination: `used_percentage: 0` with a **non-empty, real** `current_usage`. In that case:

- `isSuspiciousZero()` returns `false` (it requires `current_usage` to be all-zero), so the cache-restore branch never fires — correctly, since there's nothing to restore, the live number is genuinely more current than the cache.
- `hasGoodContext()` also returns `false` (raw `used_percentage` is `0`), so the cache-refresh branch never fires either — incorrectly, since `current_usage` clearly has a usable snapshot in it.

Nothing in `applyContextWindowFallback` (`src/context-cache.ts:293+`) reconciles the two: it only touches `used_percentage` inside the `isSuspiciousZero` branch (for the post-`/compact` transition, via `compactHint.lastCompactPostTokens`) and inside the cache-restore path. This particular combination — zero native percent, real usage — falls through both untouched.

## Minimal reproduction

Point `CLAUDE_CONFIG_DIR` at a scratch directory so this doesn't touch a real config, then feed the HUD two frames for the same `transcript_path`.

**Positive control — confirm the write path works at all.** A frame with a normal non-zero `used_percentage` writes the cache as expected:

```bash
export CLAUDE_CONFIG_DIR=/tmp/hud-repro
rm -rf "$CLAUDE_CONFIG_DIR"

echo '{
  "transcript_path": "/tmp/session-repro.jsonl",
  "model": {"display_name": "Opus"},
  "context_window": {
    "used_percentage": 38,
    "remaining_percentage": 62,
    "context_window_size": 200000,
    "current_usage": {
      "input_tokens": 74000,
      "output_tokens": 1200,
      "cache_creation_input_tokens": 1000,
      "cache_read_input_tokens": 1000
    }
  }
}' | node dist/index.js

HASH=$(node -e "console.log(require('crypto').createHash('sha256').update(require('path').resolve('/tmp/session-repro.jsonl')).digest('hex'))")
cat "$CLAUDE_CONFIG_DIR/plugins/claude-hud/context-cache/$HASH.json"
# → {"used_percentage":38, ...}   -- cache written, as expected.
```

**The bug.** Now send a second frame for the *same* `transcript_path`, simulating the in-flight-request gap: `used_percentage: 0`, but `current_usage` already shows real, higher token counts (statusline should read ~61%):

```bash
echo '{
  "transcript_path": "/tmp/session-repro.jsonl",
  "model": {"display_name": "Opus"},
  "context_window": {
    "used_percentage": 0,
    "remaining_percentage": 100,
    "context_window_size": 200000,
    "current_usage": {
      "input_tokens": 120000,
      "output_tokens": 4000,
      "cache_creation_input_tokens": 1500,
      "cache_read_input_tokens": 500
    }
  }
}' | node dist/index.js
# stdout: the context bar shows ~61% -- correct, computed live from current_usage.

cat "$CLAUDE_CONFIG_DIR/plugins/claude-hud/context-cache/$HASH.json"
# → still {"used_percentage":38, ...}  -- UNCHANGED. The renderer showed 61%,
#   the cache still says 38%, and it will keep saying 38% for as long as
#   native used_percentage keeps reading 0, no matter how much real usage
#   current_usage accumulates in the meantime.
```

## Observed vs expected

- **Observed:** the rendered line and the persisted `context-cache/<sha>.json` snapshot silently diverge on any tick where native `used_percentage` reads `0` alongside a non-empty `current_usage`, and the divergence only grows from there.
- **Expected:** the cache should hold the same value the renderer is showing at that moment — either write the fallback-derived percentage, or don't compute one at all, but the two should never quietly disagree.

## Version / commit

- Vendored/observed at commit `939eb66485832dead1b0a28a954f76f7aa2bdb06` (main, 2 commits past `v0.8.0`).
- `src/context-cache.ts` and `src/stdin.ts` are unchanged back through `v0.6.0` for the functions involved (`getNativePercent`, `isSuspiciousZero`, `hasGoodContext`, `writeCache`, `applyContextWindowFallback`), so this reproduces on any version in that range too.

## Suggested fixes

**Option A (recommended) — synthesize `used_percentage` before the write gate runs, in place.** In `applyContextWindowFallback`, add a branch alongside the existing `isSuspiciousZero` handling: when `used_percentage` is `0`/missing but `current_usage` is *not* all-zero, recompute `used_percentage` (and `remaining_percentage`) from `current_usage` right there — the same way `getContextPercent`'s fallback and the existing post-compact `postTokens` synthesis (`src/context-cache.ts:326-332`) already do — **before** `hasGoodContext()` is evaluated. This makes the mutated frame the single source of truth: `hasGoodContext()` now sees a real percentage and writes it, and since this branch runs before `render()` consumes the same `stdin` object, `getNativePercent()` picks up the synced value directly too. One code path, no possibility of the two disagreeing again. Smallest diff; reuses a pattern already in the file; ideally reuses `getTotalTokens()` from `stdin.ts` (input + cache-creation + cache-read tokens, matching the render fallback's exact formula) rather than re-deriving the sum.

**Option B (alternative) — decouple the write gate from raw `used_percentage`.** Change `hasGoodContext()` to accept "usage is present" (`!isAllUsageZero(contextWindow.current_usage)`) as sufficient, independent of the raw `used_percentage` value, and have `writeCache()` store the same fallback-derived percentage the renderer computes. This avoids mutating the frame that gets passed on to rendering, at the cost of a slightly larger diff (the synthesis logic has to live in the write path instead of a single shared mutation point).

Either way, one existing test currently encodes the buggy behavior as expected and needs its assertions updated: `applyContextWindowFallback keeps live usage when zero-percent frame already has current_usage data` (`tests/context-cache.test.js`) asserts `used_percentage` stays `0` in exactly the scenario this issue describes — that assertion is checking that the frame isn't touched, not that the cache stays correct, and it's the reason this combination fell through unnoticed.

I'm attaching a PR with option A implemented, including updated/added tests.

yotamleo:
## Summary

Fixes #742 — the context cache freezes while the statusline moves on.

When Claude Code emits `used_percentage: 0` on a tick where `current_usage` already holds real, non-zero tokens (the in-flight-request gap documented on `getNativePercent()`), the renderer correctly falls back to a percentage computed from `current_usage`, but `hasGoodContext()` gates the *cache write* on the raw `used_percentage`. The write is skipped, so the session-scoped cache stays frozen at whatever it last held and silently diverges from what is on screen.

The fix is on the write gate only: when a frame has `used_percentage: 0` but live `current_usage`, cache a percentage synthesized from `current_usage` instead of skipping the write. **The live frame is not modified.**

## Prior art

This is the cache half of an asymmetry whose render half already landed: #430 introduced the `getNativePercent()`/fallback split in `stdin.ts`, so the renderer learned to cope with a zero `used_percentage` while the cache write gate did not. #508 and #576 (and the `isSuspiciousZero()` logic they shaped) cover `used_percentage: 0` combined with an **all-zero** `current_usage` — the opposite combination from this one, which needs `current_usage` to be real and non-zero. I found no open or closed issue/PR covering the write-gate-vs-render asymmetry itself.

## Why the frame must not be mutated

An earlier revision of this PR synthesized the value *into* the frame, before `hasGoodContext()`. That was wrong, and this PR's own CI caught it — worth writing down, because the trap is not obvious.

`used_percentage` is not just a value, it is the sentinel for "a native percentage exists". `getNativePercent()` returns `null` when it is absent or `0`, and `getBufferedPercent()` applies its usage-scaled autocompact buffer **only** on that non-native fallback path. So writing a synthesized number into the frame flips the renderer out of its buffered path.

Concretely, with the `tests/integration.test.js` fixture (`context_window_size: 200000`, `current_usage.input_tokens: 45000`, no `used_percentage`):

- unmutated: no native value, so the buffered fallback runs — `scale = (0.225 - 0.05) / 0.45 = 0.3889`, `round((45000 + buffer) / 200000 * 100)` = **29%**
- mutated: `used_percentage = round(45000 / 200000 * 100) = 23`, so `getNativePercent()` returns 23, `getBufferedPercent()` short-circuits and the buffer is never applied = **23%**

The bar visibly dropped on exactly the frames the fix was meant to help. Caching a copy leaves the renderer's sentinel intact.

The raw, unbuffered percentage is the right thing to *cache*: `applyCachedContext()` restores cached values straight into `used_percentage`, where they are read back through `getNativePercent()` — so the cache holds native-equivalent percentages, not buffered display ones.

## Changes

- **`src/context-cache.ts`**
  - Import `getTotalTokens` from `./stdin.js` — the same summation the renderer's own fallback uses (input + cache-creation + cache-read, excluding output tokens).
  - `isZeroPercentWithLiveUsage()`: true when `used_percentage` is `0`/missing but `current_usage` is not all-zero. Distinct from the existing `isSuspiciousZero()`, which requires `current_usage` to *also* be all-zero — the two are mutually exclusive.
  - `synthesizePercentFromUsage()`: recomputes a percentage from `current_usage` via `getTotalTokens()`, the same formula as `getContextPercent()`'s manual fallback.
  - `applyContextWindowFallback()`: after the existing `isSuspiciousZero` handling, when the frame is zero-percent-with-live-usage, write a shallow **copy** carrying the synthesized `used_percentage`/`remaining_percentage`. The frame itself is untouched, so the renderer still takes its buffered path.
- **`tests/context-cache.test.js`**
  - The existing zero-percent-with-live-usage test now asserts the contract directly: the frame is unchanged (`used_percentage` stays `0`) while the cache file reads the synthesized `24`.
  - New test reproducing the reported shape: a cache frozen at 38% is refreshed to 61% by a later zero-percent frame carrying real `current_usage`, and that frame is likewise not mutated.

## Validation

`npm run build` (tsc) clean; `node --test tests/context-cache.test.js` and `tests/integration.test.js` both pass, including the 29% render assertion that caught the earlier approach.

One unrelated failure remains on CI here: `tests/cost-coverage.test.js` — "estimateSessionCost prices Claude 5 point releases like their base model". It is not caused by this PR. That test calls `estimateSessionCost` without pinning `now`, so it reads the wall clock, and the introductory-pricing window at `src/cost.ts:29` closed on 2026-09-01 — it fails on a clean checkout of `main` too.

I have sent that as its own one-line fix in #744, which is green. Once #744 lands I will rebase this branch onto it and CI here should follow. I have deliberately not patched around it inside this PR, since it has no overlap with `context-cache.ts` or `stdin.ts`.

### DIFF
--- a/src/context-cache.ts
+++ b/src/context-cache.ts
@@ -367,20 +367,29 @@ export function applyContextWindowFallback(
     } else if (cached) {
       applyCachedContext(contextWindow, cached);
     }
-  } else if (isZeroPercentWithLiveUsage(contextWindow)) {
-    // Zero native percent, but current_usage is real: synthesize the percent
-    // in place, before hasGoodContext() runs below, so the write gate sees it
-    // and the cached value matches what the renderer's own fallback already
-    // computed and displayed for this same frame.
-    const synced = synthesizePercentFromUsage(stdin, contextWindow);
-    if (synced !== null) {
-      contextWindow.used_percentage = synced;
-      contextWindow.remaining_percentage = 100 - synced;
-    }
   }
 
-  if (hasGoodContext(contextWindow)) {
-    writeCache(homeDir, transcriptPath, contextWindow, now, sessionName);
+  // Zero native percent, but current_usage is real: synthesize the percent
+  // that would be cached, without mutating the live frame. used_percentage
+  // doubles as getNativePercent()'s sentinel for "a native percentage
+  // exists" (see stdin.ts), so writing it here would make the renderer skip
+  // getBufferedPercent()'s autocompact buffer for this same frame.
+  const syntheticPercent = isZeroPercentWithLiveUsage(contextWindow)
+    ? synthesizePercentFromUsage(stdin, contextWindow)
+    : null;
+
+  const frameToCache = hasGoodContext(contextWindow)
+    ? contextWindow
+    : syntheticPercent !== null
+      ? {
+          ...contextWindow,
+          used_percentage: syntheticPercent,
+          remaining_percentage: 100 - syntheticPercent,
+        }
+      : null;
+
+  if (frameToCache) {
+    writeCache(homeDir, transcriptPath, frameToCache, now, sessionName);
     if (deps.random() < SWEEP_SAMPLE_RATE) {
       sweepCacheDir(getCacheDir(homeDir), now);
     }
--- a/src/context-cache.ts
+++ b/src/context-cache.ts
@@ -4,6 +4,7 @@ import * as path from "node:path";
 import { createHash } from "node:crypto";
 import { getHudPluginDir } from "./claude-config-dir.js";
 import { createDebug } from "./debug.js";
+import { getTotalTokens } from "./stdin.js";
 import type { StdinData } from "./types.js";
 
 const debug = createDebug('context-cache');
@@ -255,6 +256,39 @@ function hasGoodContext(contextWindow: ContextWindow): boolean {
 }
 
 /**
+ * True when the native percentage reads zero but current_usage already has
+ * real tokens in it — the in-flight-request gap documented on
+ * getNativePercent() in stdin.ts (Claude Code can emit used_percentage: 0
+ * before the first API response arrives while current_usage already reflects
+ * the initial-context tokens). Distinct from isSuspiciousZero(), which only
+ * fires when current_usage is ALSO all zero.
+ */
+function isZeroPercentWithLiveUsage(contextWindow: ContextWindow): boolean {
+  const usedPercentage = contextWindow.used_percentage ?? 0;
+  if (usedPercentage !== 0) {
+    return false;
+  }
+  return !isAllUsageZero(contextWindow.current_usage);
+}
+
+/**
+ * Recompute used_percentage from current_usage the same way stdin.ts's
+ * getContextPercent()/getBufferedPercent() fallback does (via
+ * getTotalTokens()), so the value we cache matches what the renderer is
+ * already showing. Returns null when context_window_size is unusable.
+ */
+function synthesizePercentFromUsage(
+  stdin: StdinData,
+  contextWindow: ContextWindow
+): number | null {
+  const size = contextWindow.context_window_size ?? 0;
+  if (size <= 0) {
+    return null;
+  }
+  return Math.min(100, Math.round((getTotalTokens(stdin) / size) * 100));
+}
+
+/**
  * Merge cached context fields into the current frame.
  * Prefer the frame's context_window_size when already present.
  */
@@ -332,6 +366,16 @@ export function applyContextWindowFallback(
       }
     } else if (cached) {
       applyCachedContext(contextWindow, cached);
+    }
+  } else if (isZeroPercentWithLiveUsage(contextWindow)) {
+    // Zero native percent, but current_usage is real: synthesize the percent
+    // in place, before hasGoodContext() runs below, so the write gate sees it
+    // and the cached value matches what the renderer's own fallback already
+    // computed and displayed for this same frame.
+    const synced = synthesizePercentFromUsage(stdin, contextWindow);
+    if (synced !== null) {
+      contextWindow.used_percentage = synced;
+      contextWindow.remaining_percentage = 100 - synced;
     }
   }
 
--- a/tests/context-cache.test.js
+++ b/tests/context-cache.test.js
@@ -175,7 +175,7 @@ test('applyContextWindowFallback ignores corrupted cache without throwing', asyn
   }
 });
 
-test('applyContextWindowFallback synthesizes used_percentage when a zero-percent frame already has current_usage data', async () => {
+test('applyContextWindowFallback caches a synthesized percent without mutating a zero-percent frame that has live current_usage', async () => {
   const tempHome = await createTempHome();
   const transcriptPath = '/tmp/session-a.jsonl';
 
@@ -199,9 +199,12 @@ test('applyContextWindowFallback synthesizes used_percentage when a zero-percent
 
     applyContextWindowFallback(stdin, makeDeps(tempHome, 1_100_000));
 
-    // getTotalTokens excludes output_tokens: (48000 + 400 + 100) / 200000 = 24%.
-    assert.equal(stdin.context_window.used_percentage, 24);
-    assert.equal(stdin.context_window.remaining_percentage, 76);
+    // The live frame's used_percentage is the renderer's sentinel for "a
+    // native percentage exists" (getNativePercent() in stdin.ts): writing to
+    // it here would disable getBufferedPercent()'s autocompact buffer. The
+    // frame must stay untouched even though a percent gets cached.
+    assert.equal(stdin.context_window.used_percentage, 0);
+    assert.equal(stdin.context_window.remaining_percentage, 100);
     assert.deepEqual(stdin.context_window.current_usage, {
       input_tokens: 48000,
       output_tokens: 2000,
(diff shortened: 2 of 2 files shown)


yotamleo:
Pushed `d491b1e` — CI here is now 1111 passing with a single failure, `estimateSessionCost prices Claude 5 point releases like their base model`, which is unrelated to this PR: it reads the wall clock and started failing on `main` when the introductory-pricing window closed on 2026-09-01. Sent as its own one-line fix in #744 (green); this PR should follow once that lands, and I've left it unpatched here rather than working around it.

The earlier red was mine and is fixed. Details are in the updated description, but the short version is worth stating inline: `used_percentage` doubles as `getNativePercent()`'s sentinel, so synthesizing a value *into* the frame silently took the renderer off `getBufferedPercent()`'s buffered path and moved the bar 29% → 23%. The fix now caches a shallow copy and leaves the frame untouched.

PR #751, opened 2026-09-04
fix: show the running session's Claude Code version, not the newest installed one

foobar:
## Summary

`display.showClaudeCodeVersion` shows the version of whatever binary `claude` on PATH currently resolves to — not the version the session process is actually executing. After Claude Code auto-updates, every session's HUD instantly "upgrades", including sessions still running the old binary (repro: session running 2.1.252, confirmed via `readlink /proc/<pid>/exe`, shows `CC v2.1.260`).

Root cause: `getClaudeCodeVersion()` (`src/version.ts`) resolves `claude` by scanning PATH and executes `claude --version`; the statusline input JSON is never consulted. That input already carries `version`, taken from the running process's own build-time VERSION constant — a documented field of the [statusline input JSON](https://code.claude.com/docs/en/statusline) (e.g. `"version": "2.1.90"`). `StdinData` just didn't declare it.

Fix: prefer the self-reported version, but only after it validates as a single version token — stdin is untrusted, so the new `parseReportedClaudeCodeVersion()` accessor (`src/version.ts`) requires the whole trimmed value to match the version pattern (anchored, 64-char cap); anything else falls back to the verified PATH lookup:

```ts
const claudeCodeVersion = config.display.showClaudeCodeVersion
    ? (parseReportedClaudeCodeVersion(stdin.version) ?? await deps.getClaudeCodeVersion())
    : undefined;
```

- `src/version.ts`: new `parseReportedClaudeCodeVersion()` accessor (shares the version-token pattern with the `claude --version` parser)
- `src/types.ts`: declare `version?: string` on `StdinData`
- `src/index.ts`: prefer the validated `stdin.version`, fall back to `deps.getClaudeCodeVersion()`
- `README.md` / `README.zh.md`: document the running-version semantics

One residual worth naming: a third-party statusline feeder that puts its own semver in `version` (e.g. wrappers that construct the stdin JSON themselves, like the Claudian Obsidian plugin referenced in `src/index.ts`) would have it displayed as the CC version. Validation requires an exactly version-shaped string but can't tell whose version it is — if you'd rather keep explicit semantics, happy to rework as a `display.versionSource: "running" | "installed"` option.

Possible follow-up (not in this diff): transcript JSONL entries already carry the running process's version per entry, and it steps mid-session on auto-update. Chaining `stdin.version` → `transcript.version` → PATH would also cover sessions started on ≤2.1.114 builds (which never send `stdin.version`) that auto-update mid-session. Happy to do that as a separate PR if you want it.

## Testing

- [x] `npm test` — 1122 tests, 1115 pass. The 1 failure ("estimateSessionCost prices Claude 5 point releases like their base model", `tests/cost-coverage.test.js`) is pre-existing on `main` and unrelated to this PR: it asserts Sonnet 5.1 costs $2/M input without pinning `now`, but Sonnet 5's introductory pricing ended 2026-09-01 UTC (`SONNET_5_PROMO_END_MS`, `src/cost.ts:29`), so the unpinned expectation evaluates to $3 now. Its sibling test pins `now` explicitly — doing the same there would fix it.
- [x] `npm run test:coverage` — 94.2% statements / 87.4% branches / 94.2% lines overall (same 1 pre-existing failure)

Tests written first (each watched failing against the old behavior): a unit matrix for `parseReportedClaudeCodeVersion` in `tests/version.test.js` (clean/whitespace-padded versions pass; prose like "Opus 4.6", escape-polluted strings, non-strings, blanks, and overlong values fall back), and wiring tests in `tests/index.test.js` (stdin's version wins and the PATH lookup is skipped; junk, non-string, and empty values fall back to PATH resolution).

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,7 +243,7 @@ Simplified and Traditional Chinese HUD labels are available as explicit opt-ins.
 | `display.showCompactions` | boolean | false | Show how many context compactions (manual `/compact` or auto) have occurred this session, counted from transcript `compact_boundary` entries, e.g. `Compactions: 2`. Hidden until the first compaction |
 | `display.showEffortLevel` | boolean | false | Show the current reasoning effort in the model badge. Ultracode renders as `ultracode(xhigh)`, detected from the session transcript so it tracks `/effort` changes made at runtime |
 | `display.effortFormat` | `full` \| `symbol` \| `text` | `full` | How the effort renders when `display.showEffortLevel` is on: symbol and level text (`◑ high`), symbol only (`◑`), or level text only (`high`). Ultracode keeps the full `◕ ultracode(xhigh)` form under `symbol` so the marker is not lost, and levels without a known symbol fall back to the level text |
-| `display.showClaudeCodeVersion` | boolean | false | Show the installed Claude Code version, e.g. `CC v2.1.81` |
+| `display.showClaudeCodeVersion` | boolean | false | Show the running session's Claude Code version, e.g. `CC v2.1.81`. Falls back to the installed version when the session doesn't report a usable one |
 | `display.showMemoryUsage` | boolean | false | Show an approximate system RAM usage line in expanded layout |
 | `display.showPromptCache` | boolean | false | Show the wall-clock time the session's prompt cache expires, read from the transcript |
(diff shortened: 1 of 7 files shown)


foobar:
@jarrodwatts PTAL

